PR #121, opened 2015-10-02
Migrate to OkHttp

hzalaz:


steffandroid:
Oh wow, amazing work, thank you for this.

Issue #126, opened 2015-11-10
Add ImageView to background on login/reset password screens


xorgate:
Our client wants the login page to be customized in a way that a bitmap is shown behind the login controls. Currently I can only set that bitmap on the LinearLayout, but it has only rudimentary support for it.

I think it would be a good idea to wrap this LinearLayout in a FrameLayout, which also has an ImageView in it. This way we can properly set a background bitmap and have it scale the way we desire.

xorgate:
Fixed in https://github.com/auth0/Lock.Android/pull/128

PR #128, opened 2015-11-10
add ImageView and styles to activity to have more control over backgr…

xorgate:
…ound image

As per issue #126


hzalaz:
Closes #126

PR #133, opened 2015-11-11
Update the minimum api level to 15

nikolaseu:


ritdaw:
What has been added to jump from 14 > 15?

hzalaz:
@ritdaw Actually FB SDK requires 15 and though we could get a way around it we think it's not worth the trouble since there is are very few devices with that version (it doesn't even appear in http://developer.android.com/about/dashboards/index.html) and Google recommended that 4.0.3 (api 15) be the default ICS for manufacturers.

ritdaw:
Thanks, just thought I'd ask the question. You are totally right, only just realised that we have 0 users using our app on SDK 14, they are all 15.

Issue #140, opened 2015-11-26
Support customizing "user is blocked" message


hzalaz:
The response is 
```
{
  "error": "unauthorized",
  "error_description": "user is blocked"
}
```

nikolaseu:
Done in #149

PR #149, opened 2015-12-09
Feature: Add option to customize the "user is blocked" message

nikolaseu:
In cases where the response is:

```json
{
  "error": "unauthorized",
  "error_description": "user is blocked"
}
```

We don't use the 'error_description', instead we use a string resource that can be customized/translated

Closes issue #140


hzalaz:
Sadly we need to check also for the `user is blocked` message

PR #159, opened 2015-12-21
Add option to force webview on adfs connections

### LINKED ISSUE #158: Add an option to use Webview with ADFS connections

nikolaseu:
To force the use of webview on certain connections the user can specify the connections in the Lock.Builder

```java
    new Lock.Builder()
        .loadFromApplication(this)
        .closable(true)
        .useEmail(true)
        /* other options */
        .enterpriseConnectionsUsingWebForm("connection1", "connection2")

```

Closes #158

### DIFF
--- a/lock/src/main/java/com/auth0/lock/Lock.java
+++ b/lock/src/main/java/com/auth0/lock/Lock.java
@@ -114,6 +114,7 @@ public class Lock {
 
     private Configuration configuration;
     private List<String> connections;
+    private List<String> enterpriseConnectionsUsingWebForm;
     private String defaultDatabaseConnection;
     private boolean signUpEnabled;
     private boolean changePasswordEnabled;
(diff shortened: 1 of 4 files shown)


lock/src/main/java/com/auth0/lock/Lock.java
```diff
@@ -362,6 +367,7 @@ public static Lock getLock(Activity activity) {
         private boolean useEmail;
         private String defaultDBConnectionName;
         private List<String> connections;
+        private List<String> enterpriseConnectionsUsingWebForm;
```
hzalaz: Let's initialise this with empty List and enforce it to be non-null

hzalaz: So we can avoid https://github.com/auth0/Lock.Android/pull/159/files#diff-035023a6fd0804236699900c41690be7R258

PR #192, opened 2016-01-29
Better social ui

lbalmaceda:
remake the social buttons layout with custom colors, texts and icons. Based on the following screenshot and v1 resources.
based on this


lock/src/main/java/com/auth0/android/lock/views/SocialButton.java
```diff
@@ -1,50 +1,142 @@
...
+        super(context, attrs, defStyleAttr);
+        this.configured = false;
+        init();
+    }
+
+    @SuppressWarnings("deprecation")
     private void init() {
-        setClickable(true);
-        setFocusable(true);
+        inflate(getContext(), R.layout.com_auth0_btn_social_large, this);
```
hzalaz: missed `lock` in the resource prefix

lock/src/main/java/com/auth0/android/lock/views/SocialButtonConfig.java
```diff
@@ -0,0 +1,98 @@
...
+import android.support.annotation.DrawableRes;
+import android.support.annotation.NonNull;
+import android.support.annotation.StringRes;
+import android.util.Log;
+
+import com.auth0.android.lock.R;
+import com.auth0.android.lock.utils.Strategies;
+import com.auth0.android.lock.utils.Strategy;
+
+class SocialButtonConfig {
```
hzalaz: Let's add whatever tests we can, this is a probably a point of failure/crashes if we mess up

hzalaz: Also not a big fan of the name but I can't think of other one (maybe `SocialConfig` or `SocialResourcesProvider`) and since it's private maybe its not a big deal

lock/src/main/java/com/auth0/android/lock/views/SocialButtonConfig.java
```diff
@@ -0,0 +1,98 @@
...
+    private static final String ICON_RESOURCE_FORMAT = "com_auth0_social_icon_%s";
```
hzalaz: Let's prefix all with the same one `com_auth0_lock`

lock/src/main/java/com/auth0/android/lock/views/ViewUtils.java
```diff
@@ -0,0 +1,35 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.auth0.android.lock.views;
+
+import android.content.res.Resources;
+import android.util.TypedValue;
+
+class ViewUtils {
```
hzalaz: make it abstract

PR #193, opened 2016-02-01
Header widget

lbalmaceda:
Login header customizable from the app's theme, defaults to Auth0 colors and logo.


lbalmaceda:
before merging this PR, I need the Auth0 logo resource.

lock/src/main/java/com/auth0/android/lock/views/HeaderView.java
```diff
@@ -0,0 +1,102 @@
...
+        super(context, attrs);
+        init(attrs);
+    }
+
+    public HeaderView(Context context, AttributeSet attrs, int defStyleAttr) {
+        super(context, attrs, defStyleAttr);
+        init(attrs);
+    }
+
+    @SuppressWarnings("deprecation")
```
hzalaz:
```
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
    return resources.getDrawable(id, context.getTheme());
} else {
    return resources.getDrawable(id);
}
```

Use that, it won't avoid the warning but it will be the proper way

PR #194, opened 2016-02-01
handle network errors on webview flow

lbalmaceda:
Check for network status before using the webview. Also let the user retry if there was an error. Note: for this to work, we now require ACCESS_NETWORK_STATE permission.

### DIFF
--- a/lock/src/main/java/com/auth0/android/lock/provider/WebViewActivity.java
+++ b/lock/src/main/java/com/auth0/android/lock/provider/WebViewActivity.java
@@ -24,12 +24,14 @@
 
 package com.auth0.android.lock.provider;
 
+import android.annotation.TargetApi;
 import android.content.Context;
 import android.content.Intent;
 import android.graphics.Bitmap;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Bundle;
 import android.support.v7.app.ActionBar;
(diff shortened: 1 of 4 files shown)


hzalaz:
Looks good, can we allow this to work when the permission is not added @lbalmaceda ?

hzalaz:
Also let's have a "Oops" view too, for an error we don't understand (design pending)

lock/src/main/java/com/auth0/android/lock/provider/WebViewActivity.java
```diff
@@ -60,15 +75,44 @@ protected void onCreate(Bundle savedInstanceState) {
...
         progressBar = (ProgressBar) findViewById(R.id.com_auth0_lock_progressbar);
+        progressBar.setIndeterminate(true);
+        progressBar.setMax(100);
+        errorView = findViewById(R.id.com_auth0_lock_error_view);
+        errorMessage = (TextView) findViewById(R.id.com_auth0_lock_text);
+        Button retryButton = (Button) findViewById(R.id.com_auth0_lock_retry);
+        retryButton.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                isShowingError = false;
```
hzalaz: Can we replace this by checking the visibility of `errorMessage`?

lbalmaceda: sure!

PR #196, opened 2016-02-04
Db-connection: signup

lbalmaceda:
Signup flow for the DbConnection.

### DIFF
--- a/lock/src/main/java/com/auth0/android/lock/enums/UsernameStyle.java
+++ b/lock/src/main/java/com/auth0/android/lock/enums/UsernameStyle.java
@@ -30,5 +30,5 @@ package com.auth0.android.lock.enums;
 public enum UsernameStyle {
     EMAIL,
     USERNAME,
-    USERNAME_OR_EMAIL
+    DEFAULT
 }
--- a/lock/src/main/java/com/auth0/android/lock/enums/UsernameStyle.java
+++ b/lock/src/main/java/com/auth0/android/lock/enums/UsernameStyle.java
@@ -30,5 +30,5 @@ package com.auth0.android.lock.enums;
 public enum UsernameStyle {
     EMAIL,
     USERNAME,
-    DEFAULT
+    USERNAME_OR_EMAIL
 }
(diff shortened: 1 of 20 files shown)


lock/src/main/java/com/auth0/android/lock/views/DbLayout.java
```diff
@@ -0,0 +1,157 @@
...
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.Button;
+import android.widget.RelativeLayout;
+
+import com.auth0.android.lock.Configuration;
+import com.auth0.android.lock.R;
+import com.squareup.otto.Bus;
+
+public class DbLayout extends RelativeLayout implements View.OnClickListener {
```
hzalaz: Let's use full names like `DatabaseLayout`

lock/src/main/java/com/auth0/android/lock/enums/UsernameStyle.java
```diff
@@ -30,5 +30,5 @@
 public enum UsernameStyle {
     EMAIL,
     USERNAME,
-    DEFAULT
+    USERNAME_OR_EMAIL
```
hzalaz: Let's stick with `DEFAULT`, or a better name, since it can be user or email or user and email depending on the connection configuration. Other option is to have `null` mean default

lock/src/main/java/com/auth0/android/lock/events/DbSignUpEvent.java
```diff
@@ -25,22 +25,42 @@
 package com.auth0.android.lock.events;
 
 
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+
 public class DbSignUpEvent {
```
hzalaz: Same as `DBLayout`

lock/src/main/java/com/auth0/android/lock/views/LoginFormView.java
```diff
@@ -25,70 +25,71 @@
...
+        passwordInput.setDataType(ValidatedInputView.DataType.PASSWORD);
+
+        switch (configuration.getUsernameStyle()) {
+            case EMAIL:
+                usernameEmailInput.setDataType(ValidatedInputView.DataType.EMAIL);
+                break;
+            case USERNAME:
+                usernameEmailInput.setDataType(ValidatedInputView.DataType.USERNAME);
+                break;
+            default:
```
hzalaz: why default if the enum only has 3 members?

PR #200, opened 2016-02-11
Enterprise login form and authentication methods

lbalmaceda:


lock/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -342,6 +351,43 @@ public void onDatabaseAuthenticationRequest(DatabaseChangePasswordEvent event) {
...
+        if (options == null) {
+            return;
+        } else if (event.getConnectionName().equals(Strategies.ActiveDirectory.getName()) && configuration.getDefaultActiveDirectoryConnection() == null) {
+            return;
+        } else if (configuration.getEnterpriseStrategies().isEmpty()) {
+            return;
+        }
+
+        progress.showProgress();
+        AuthenticationAPIClient apiClient = new AuthenticationAPIClient(options.getAccount());
```
hzalaz: It's time we ask LockOptions for the `AuthenticationAPIClient`

lock/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -342,6 +351,43 @@ public void onDatabaseAuthenticationRequest(DatabaseChangePasswordEvent event) {
...
+        apiClient.setDefaultDbConnection(event.getConnectionName());
```
hzalaz: We should not do this, we should instead set the `connection` entry in the parameters of the request.

lock/src/main/java/com/auth0/android/lock/views/DomainFormView.java
```diff
@@ -0,0 +1,243 @@
...
+        return valid;
+    }
+
+    @Nullable
+    private Strategy searchDomain(String domain) {
+        if (domain.isEmpty()) {
+            return null;
+        }
+        for (Strategy s : filteredEnterpriseStrategies) {
+            if (s.getType() == Strategies.Type.ENTERPRISE && domain.toLowerCase().contains(s.getName())) {
```
hzalaz: This is not right, the domains are contained in the connection values, the name of the strategy is something different

lock/src/main/java/com/auth0/android/lock/views/DomainFormView.java
```diff
@@ -0,0 +1,243 @@
...
+
+            }
+
+            @Override
+            public void onTextChanged(CharSequence s, int start, int before, int count) {
+            }
+
+            @Override
+            public void afterTextChanged(Editable s) {
+                String text = s.toString();
```
hzalaz: Let's move the logic of domain matching to a different object and properly test it (otherwise it won't be tested at all). Maybe an object that for a domain returns the matched connection or nothing

lock/src/main/java/com/auth0/android/lock/views/DomainFormView.java
hzalaz: Also we need to return the connection not the strategy, since an enterprise strategy might hold more than one connection with different domains

lock/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -342,6 +351,43 @@ public void onDatabaseAuthenticationRequest(DatabaseChangePasswordEvent event) {
...
+        CallbackHelper helper = new CallbackHelper(pkgName);
+        lastIdp = new WebIdentityProvider(helper, options.getAccount(), idpCallback);
+        lastIdp.setUseBrowser(options.useBrowser());
+        lastIdp.setParameters(options.getAuthenticationParameters());
+        lastIdp.start(LockActivity.this, event.getConnectionName());
+    }
+
+    @SuppressWarnings("unused")
+    @Subscribe
+    public void onEnterpriseAuthenticationRequest(EnterpriseROLoginEvent event) {
```
hzalaz: I'm not buying separating the enterprise login with 2 different event. I'd rather make the decision in a single method (by fetching the ro enabled flag from the event), also it will make things easier when we add an option to go always with web flow.

lock/src/main/java/com/auth0/android/lock/views/DomainFormView.java
```diff
@@ -0,0 +1,243 @@
...
+                if (currentStrategy != null) {
+                    actionButton.setEnabled(true);
+                    actionButton.setText(String.format(getResources().getString(R.string.com_auth0_lock_action_login_with), currentStrategy.getName()));
+                } else {
+                    resetDomain();
+                }
+            }
+        });
+
+        switch (configuration.getUsernameStyle()) {
```
hzalaz: I bet we have this code copy pasted several times, shouldn't we move this to the input class?

lock/src/main/java/com/auth0/android/lock/views/EnterpriseLayout.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    private void init() {
+        showDomainForm();
+    }
+
+    @Override
+    public void onClick(View v) {
+        switch (v.getId()) {
+            case GO_BACK_BTN_ID:
```
hzalaz: Where is this value set in a button?

lock/src/main/java/com/auth0/android/lock/views/EnterpriseLayout.java
```diff
@@ -0,0 +1,84 @@
...
+                showDomainForm();
+                break;
+        }
+    }
+
+    private void showDomainForm() {
+        this.removeView(findViewById(FORM_ID));
+
+        DomainFormView domainForm = new DomainFormView(getContext(), this.bus, this.configuration);
+        domainForm.setId(FORM_ID);
```
hzalaz: Not sure if this is safe to use, since we are hardcoding the value and not letting android generate it. Also we should probably keep the form reference instead

lock/src/main/java/com/auth0/android/lock/utils/EmailParser.java
```diff
@@ -0,0 +1,150 @@
...
+import com.auth0.Auth0Exception;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Helper class to find out which email domains can be valid for the current Auth0 configuration.
+ */
+public class EmailParser {
```
hzalaz: It's more like a `DomainMatcher`

lock/src/main/java/com/auth0/android/lock/utils/EmailParser.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private static final String DOMAIN_KEY = "domain";
+    private static final String DOMAIN_ALIASES_KEY = "domain_aliases";
+
+    private List<Strategy> strategies;
+
+    public EmailParser(List<Strategy> strategies) {
+        if (strategies == null || strategies.isEmpty()) {
+            throw new Auth0Exception("You must provide a valid list of Strategies.");
```
hzalaz: Let's avoid this kind of behaviour whenever possible since it will crash the app unless we are catching this kind of exception (which I doubt). If there is no enterprise strategies we should always return no connection.

lock/src/main/java/com/auth0/android/lock/utils/EmailParser.java
```diff
@@ -0,0 +1,150 @@
...
+
+
+    /**
+     * Tries to find the Strategy holder of the given connection.
+     *
+     * @param connection to search for.
+     * @return a Strategy if found, null otherwise.
+     */
+    @Nullable
+    public Strategy strategyForConnection(Connection connection) {
```
hzalaz: I was thinking this last night, we could inject in the `Connection` the flag if its `/ro`-able like we do with the types when we load them from cdn

lock/src/main/java/com/auth0/android/lock/utils/EnterpriseConnectionMatcher.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * Extracts the domain part from the email
+     *
+     * @param email to parse
+     * @return the domain String if found, an empty String otherwise
+     */
+
+    public String extractDomain(String email) {
```
hzalaz: why public?

lock/src/main/java/com/auth0/android/lock/utils/EnterpriseConnectionMatcher.java
```diff
@@ -0,0 +1,131 @@
...
+        return domain;
+    }
+
+    /**
+     * Extracts the username part from the email
+     *
+     * @param email to parse
+     * @return the username String if found, an empty String otherwise
+     */
+    public String extractUsername(String email) {
```
hzalaz: why public?

lock/src/main/java/com/auth0/android/lock/utils/EnterpriseConnectionMatcher.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+        int indexDot = email.indexOf(".", indexAt);
+        String domain;
+        if (indexDot == -1) {
+            domain = email.substring(indexAt);
+        } else {
+            domain = email.substring(indexAt, indexDot);
+        }
+        if (domain.isEmpty()) {
+            return "";
```
hzalaz: why not return null?

lock/src/main/java/com/auth0/android/lock/utils/EnterpriseConnectionMatcher.java
lbalmaceda: @hzalaz  To test the domain extraction it needs a public scope. Do you prefer to test just the `parse` method?

lock/src/main/java/com/auth0/android/lock/utils/EnterpriseConnectionMatcher.java
lbalmaceda: @hzalaz this method is used on the `DomainFormView` class, when pre-filling the next username/email form with the right text (full email or only the username part)

lock/src/main/java/com/auth0/android/lock/utils/EnterpriseConnectionMatcher.java
hzalaz: We should not test internal implementation details but the public behaviour of the class.

lock/src/main/java/com/auth0/android/lock/utils/EnterpriseConnectionMatcher.java
hzalaz: :+1:

PR #205, opened 2016-02-23
Passwordless Email with Link/Code authentication

lbalmaceda:


lock/src/main/java/com/auth0/android/lock/Lock.java
```diff
@@ -87,8 +88,9 @@ public Options getOptions() {
         return options;
     }
 
-    public static Builder newBuilder() {
-        return new Lock.Builder();
+    @SuppressWarnings("unused")
+    public static Builder newBuilder(@NonNull Auth0 account, @NonNull AuthenticationCallback callback) {
```
hzalaz: Docs?

lock/src/main/java/com/auth0/android/lock/Lock.java
```diff
@@ -161,29 +167,30 @@ private void processEvent(Intent eventData) {
...
+            this.callback = callback;
             options = new Options();
+            options.setAccount(account);
             options.setAuthenticationParameters(defaultParams);
         }
 
         public Lock build() {
+            if (options.getAccount() == null) {
+                Log.e(TAG, "You need to specify the com.auth0.Auth0 object with the Auth0 Account details.");
+                throw new IllegalArgumentException("Missing Auth0 account information.");
```
hzalaz: `IllegalStateException` will be more adequate

lock/src/main/java/com/auth0/android/lock/PasswordlessLock.java
```diff
@@ -0,0 +1,217 @@
...
+            this.callback = callback;
+            options = new Options();
+            options.setAccount(account);
+            options.setAuthenticationParameters(defaultParams);
+        }
+
+        public PasswordlessLock build() {
+            if (options.getAccount() == null) {
+                Log.e(TAG, "You need to specify the com.auth0.Auth0 object with the Auth0 Account details.");
+                throw new IllegalArgumentException("Missing Auth0 account information.");
```
hzalaz: `IllegalStateException` will be more adequate

lock/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
```diff
@@ -0,0 +1,236 @@
...
+        rootView = (LinearLayout) findViewById(R.id.com_auth0_lock_content);
+        initLockUI();
+    }
+
+    private boolean isLaunchConfigValid() {
+        options = getIntent().getParcelableExtra(Lock.OPTIONS_EXTRA);
+        if (options == null) {
+            Log.e(TAG, "You need to specify the com.auth0.android.lock.Options in the Lock.OPTIONS_EXTRA of the Intent for PasswordlessLockActivity to launch. " +
+                    "Use com.auth0.android.lock.PasswordlessLock.Builder to generate one.");
+            throw new IllegalArgumentException("Missing com.auth0.android.lock.Options in intent");
```
hzalaz: Never kill an activity like this

lock/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
```diff
@@ -0,0 +1,236 @@
...
+        progress.showResult("");
+        rootView = (LinearLayout) findViewById(R.id.com_auth0_lock_content);
+        initLockUI();
+    }
+
+    private boolean isLaunchConfigValid() {
+        options = getIntent().getParcelableExtra(Lock.OPTIONS_EXTRA);
+        if (options == null) {
+            Log.e(TAG, "You need to specify the com.auth0.android.lock.Options in the Lock.OPTIONS_EXTRA of the Intent for PasswordlessLockActivity to launch. " +
+                    "Use com.auth0.android.lock.PasswordlessLock.Builder to generate one.");
```
hzalaz: No need for full package name, just `PasswordlessLock.Builder` will do. Also let's clarify that they need to create the intent using the builder

hzalaz: These messages should tell the user what to do without knowing the inner workings of the lib

lock/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
```diff
@@ -0,0 +1,236 @@
...
+        processIncomingIntent(intent);
+
+        super.onNewIntent(intent);
+    }
+
+    private void processIncomingIntent(Intent intent) {
+        if (options != null && options.passwordlessMode() == PasswordlessMode.EMAIL_LINK) {
+            String code = intent.getData().getQueryParameter("code");
+            if (code == null || code.isEmpty() || lastPasswordlessEmailOrNumber == null) {
+                Log.w(TAG, "Code was invalid.");
```
hzalaz: So if we open a link that is invalid we'll leave the `PasswordlessLockActivity` alive and the user has to go back manually?.

Also what will the user see in this case?

lock/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
```diff
@@ -0,0 +1,236 @@
...
+        if (options == null) {
+            return;
+        } else if (event.getEmailOrNumber().isEmpty()) {
+            return;
+        }
+
+        progress.showProgress();
+        AuthenticationAPIClient apiClient = new AuthenticationAPIClient(options.getAccount());
+        if (event.getCode() != null) {
+            AuthenticationRequest answerRequest = null;
```
hzalaz: why `answerRequest` and not `loginRequest`?

lock/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
```diff
@@ -0,0 +1,236 @@
...
+            if (event.getMode() == PasswordlessMode.EMAIL_CODE || event.getMode() == PasswordlessMode.EMAIL_LINK) {
```
hzalaz: It will probably make sense to ask the event for the request like:

```java
AuthenticationRequest request = event.getLoginRequest(apiClient);
```

hzalaz: And we could test this logic too

lock/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
```diff
@@ -0,0 +1,236 @@
...
+            }
+            answerRequest.addParameters(options.getAuthenticationParameters());
+            answerRequest.start(authCallback);
+            return;
+        }
+
+        lastPasswordlessEmailOrNumber = event.getEmailOrNumber();
+        ParameterizableRequest<Void> codeRequest=null;
+        if (event.getMode() == PasswordlessMode.EMAIL_CODE) {
+            codeRequest = apiClient.passwordlessWithEmail(event.getEmailOrNumber(), PasswordlessType.CODE);
```
hzalaz: Same as above but for the code request

lock/src/main/java/com/auth0/android/lock/views/ValidatedInputView.java
```diff
@@ -115,6 +117,14 @@ private void setupInputValidation() {
                 inputErrorIcon = R.drawable.com_auth0_lock_ic_input_username_error;
                 hint = getResources().getString(R.string.com_auth0_lock_hint_username);
                 break;
+            case CODE:
+                input.setInputType(InputType.TYPE_NUMBER_VARIATION_NORMAL);
+                //Fixme: Code length is configured in the dashboard. How can we get this value?
```
hzalaz: We should not restrict a limit of characters since we don't know the preferred length

lock/src/main/java/com/auth0/android/lock/Lock.java
```diff
@@ -160,45 +190,85 @@ private void processEvent(Intent eventData) {
...
 
+        /**
+         * Additional Authentication parameters can be set to use with different Identity Providers.
+         *
+         * @param authenticationParameters a non-null Map containing the parameters as Key-Values
+         * @return the current builder instance
+         */
         public Builder withAuthenticationParameters(@NonNull Map<String, Object> authenticationParameters) {
             if (authenticationParameters instanceof HashMap) {
                 options.setAuthenticationParameters((HashMap<String, Object>) authenticationParameters);
```
hzalaz: why?

lock/src/main/java/com/auth0/android/lock/PasswordlessLock.java
```diff
@@ -0,0 +1,285 @@
...
+
+        /**
+         * Additional Authentication parameters can be set to use with different Identity Providers.
+         *
+         * @param authenticationParameters a non-null Map containing the parameters as Key-Values
+         * @return the current builder instance
+         */
+        public Builder withAuthenticationParameters(@NonNull Map<String, Object> authenticationParameters) {
+            if (authenticationParameters instanceof HashMap) {
+                options.setAuthenticationParameters((HashMap<String, Object>) authenticationParameters);
```
hzalaz: why the cast?

lock/src/test/java/com/auth0/android/lock/events/PasswordlessLoginEventTest.java
```diff
@@ -0,0 +1,98 @@
...
+    public void shouldSetTheCode() throws Exception {
+        PasswordlessLoginEvent event = new PasswordlessLoginEvent(PasswordlessMode.EMAIL_CODE, EMAIL, CODE);
+
+        Assert.assertThat(event.getEmailOrNumber(), is(equalTo(EMAIL)));
+        Assert.assertThat(event.getMode(), is(equalTo(PasswordlessMode.EMAIL_CODE)));
+        Assert.assertThat(event.getCode(), is(equalTo(CODE)));
+    }
+
+    @Test
+    public void shouldGetCodeRequest() throws Exception {
```
hzalaz: This test will be simpler if we use Mock and verify the correct method is called.

lock/src/test/java/com/auth0/android/lock/events/PasswordlessLoginEventTest.java
```diff
@@ -0,0 +1,98 @@
...
+        Map<String, Object> parameters = codeRequest.getParameterBuilder().asDictionary();
+
+        assertThat(codeRequest, is(not(nullValue())));
+        assertThat((String) parameters.get(EMAIL_KEY), is(equalTo(EMAIL)));
+        assertThat((String) parameters.get(SEND_KEY), is(equalTo(CODE_MODE)));
+        assertThat(parameters, IsMapContaining.hasKey(CONNECTION_KEY));
+    }
+
+    @Test
+    public void shouldGetLoginRequest() throws Exception {
```
hzalaz: If we use Mock and verify the correct method is called we can make sure we are returning a login request and not any  request

lock/src/main/java/com/auth0/android/lock/PasswordlessLock.java
lbalmaceda: `Options` class requires a HashMap (Serializable) to store/read the field when implementing `Parcelable`.

hzalaz: Yes but this is conceptually wrong since we are accepting the generic interface and then force casting to a specific subclass, even if we have the `instanceof`. The better approach is to create a `HashMap` with the supplied map

PR #212, opened 2016-02-25
Filter Passwordless connections on the Configuration

lbalmaceda:


lock/src/test/java/com/auth0/android/lock/utils/ApplicationTest.java
```diff
@@ -77,6 +79,12 @@ public void shouldReturnEnterpriseStrategies() throws Exception {
         assertThat(application.getEnterpriseStrategies().size(), equalTo(3));
     }
 
+    @Test
+    public void shouldReturnPasswordlessStrategies() throws Exception {
+        Application application = newApplicationWithStrategies(Email, SMS, Facebook, Auth0);
+        assertThat(application.getPasswordlessStrategies().size(), equalTo(2));
```
hzalaz: This test is flawed since if it returns Facebook or Auth0 it won't break.

lbalmaceda: Nice catch!

lock/src/test/java/com/auth0/android/lock/utils/ApplicationTest.java
```diff
@@ -77,6 +79,12 @@ public void shouldReturnEnterpriseStrategies() throws Exception {
         assertThat(application.getEnterpriseStrategies().size(), equalTo(3));
     }
 
+    @Test
+    public void shouldReturnPasswordlessStrategies() throws Exception {
+        Application application = newApplicationWithStrategies(Email, SMS, Facebook, Auth0, SAMLP);
+        assertThat(application.getPasswordlessStrategies().size(), equalTo(2));
```
hzalaz: Same flaw applies here again. If somehow we add a bug that make it match only two besides email or sms the test won't fail, so we need to check that the collection contains only those strategies we want

PR #213, opened 2016-02-25
Passwordless mode from Application

lbalmaceda:


lock/src/main/java/com/auth0/android/lock/enums/PasswordlessType.java
```diff
@@ -0,0 +1,29 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.auth0.android.lock.enums;
+
+public enum PasswordlessType {
```
hzalaz: why not a boolean?

lock/src/main/java/com/auth0/android/lock/enums/PasswordlessType.java
```diff
@@ -0,0 +1,29 @@
...
+    LINK, CODE
```
hzalaz: We should use a boolean for type since the decision is binary. If we want to make it easier to the user to pick we can have 2 methods in the builder for each one

lock/src/main/java/com/auth0/android/lock/Configuration.java
```diff
@@ -172,27 +183,69 @@ private Strategy filterStrategy(Strategy strategy, Set<String> connections) {
...
+            usernameRequired = getDefaultDatabaseConnection().booleanForKey(REQUIRES_USERNAME_KEY);
         }
 
-        usernameRequired = getDefaultDatabaseConnection().booleanForKey(REQUIRES_USERNAME_KEY);
+        PasswordlessType passwordlessType = options.passwordlessType();
+        if (getPasswordlessStrategies().isEmpty()) {
+            passwordlessMode = null;
+            passwordlessType = null;
+        }
+        if (passwordlessType != null) {
```
hzalaz: We should start thinking in connection names like we do for the other strategies. We should do the same auth0 strategy does, pick the first connection

lock/src/main/java/com/auth0/android/lock/Configuration.java
```diff
@@ -179,40 +183,76 @@ private Strategy filterStrategy(Strategy strategy, Set<String> connections) {
...
+    private Strategy filterPasswordlessStrategies(List<Strategy> strategies, Set<String> connections) {
+        if (strategies == null || strategies.isEmpty()) {
+            return null;
         }
-        List<Strategy> filtered = new ArrayList<>(strategies.size());
-        for (Strategy strategy : strategies) {
-            if (connections.contains(strategy.getName())) {
-                filtered.add(strategy);
+
+        if (connections.isEmpty()) {
```
hzalaz: We should improve this, it's not ideal to have so many `return`

hzalaz: Also we are are not filtering the connections of the passwordless strategy, we are just returning it as it is, so if the client has two connections `conn1` & `conn2` inside lets say `Email` strategy, and wishes to use `conn2` it won't be possible since we are not filtering the first one

lock/src/main/java/com/auth0/android/lock/Configuration.java
```diff
@@ -92,6 +95,25 @@ public Strategy getActiveDirectoryStrategy() {
         return activeDirectoryStrategy;
     }
 
+    @Nullable
+    public Strategy getDefaultPasswordlessStrategy() {
```
hzalaz: Great, but I'll make it future-proof like this

```java
if (passwordlessStrategies.isEmtpy()) {
     return null;
}

if (passwordlessStrategies.size() == 1) {
    return passwordlessStrategies.get(0);
}

Strategy default = null;

for (Strategy s : passwordlessStrategies) {
    if (s.getName().equals(Strategies.Email.getName())) {
        default = s;
        break;
    }
}

return default || passwordlessStrategies.get(0);
```

If we have more than one we pick email or whatever is the first one in case there is no email connection

PR #214, opened 2016-02-26
Use default passwordless connection

lbalmaceda:


lock/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
```diff
@@ -255,7 +257,9 @@ public void onPasswordlessAuthenticationRequest(PasswordlessLoginEvent event) {
         if (event.getCode() != null) {
             AuthenticationRequest loginRequest = event.getLoginRequest(apiClient);
             if (loginRequest != null) {
```
hzalaz: why this is needed?

lock/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
```diff
@@ -255,7 +257,9 @@ public void onPasswordlessAuthenticationRequest(PasswordlessLoginEvent event) {
...
-                loginRequest.addParameters(options.getAuthenticationParameters());
+                HashMap<String, Object> authenticationParameters = options.getAuthenticationParameters();
+                authenticationParameters.put("connection", configuration.getDefaultPasswordlessStrategy().getConnections().get(0).getName());
```
hzalaz: I believe `AuthenticationRequest` has a `setConnection` method.

hzalaz: Also let's extract to a private method the code to get the connection name

PR #215, opened 2016-02-29
Add Social to Passwordless flow

lbalmaceda:


lock/src/main/java/com/auth0/android/lock/views/PanelHolder.java
```diff
@@ -44,10 +48,19 @@ public PanelHolder(Context context) {
         configuration = null;
     }
 
-    public PanelHolder(Context context, Bus lockBus, Configuration configuration) {
+    public PanelHolder(PasswordlessLockActivity context, Bus lockBus, Configuration configuration) {
```
hzalaz: Why we have these constructors that receive a specific activity?

lock/src/main/java/com/auth0/android/lock/views/PanelHolder.java
```diff
@@ -44,10 +48,19 @@ public PanelHolder(Context context) {
...
         super(context);
         this.bus = lockBus;
         this.configuration = configuration;
+        this.usingPasswordless = true;
```
hzalaz: Couldn't we figure this out from `configuration`?

lbalmaceda: What if the dashboard has all the 4 kinds of connections? (social/enterprise/db/passwordless), then the form will display all of those, and if instantiated from a `LockActivity` instead of a `PasswordlessLockActivity`, the passwordless flow wont work because the event subscriber method is on the `PasswordlessLockActivity`.

hzalaz: Then let's decouple this object into two one for Classic mode and another for Passwordless, this will make the logic in `init()` easier to read

PR #219, opened 2016-03-07
handle M permissions on permission required providers

lbalmaceda:


identity-core/src/main/java/com/auth0/identity/AuthorizedIdentityProvider.java
```diff
@@ -0,0 +1,146 @@
...
+        return identityProvider.authorize(activity, requestCode, resultCode, data);
+    }
+
+    @Override
+    public void clearSession() {
+        identityProvider.clearSession();
+    }
+
+    /**
+     * Defines which permissions are required by this Identity Provider to work
```
hzalaz: Explain what permissions are and give an example, if I read this I don't know if its Android's or Auth0's

identity-core/src/main/java/com/auth0/identity/AuthorizedIdentityProvider.java
```diff
@@ -0,0 +1,146 @@
...
+     *
+     * @return the required permissions
+     */
+    public abstract String[] getRequiredPermissions();
+
+    /**
+     * Retry the last permission request issued to the user. If there is no recent
+     * request, this method will do nothing.
+     */
+    public void retryLastPermissionRequest() {
```
hzalaz: Do we use this or is for just in case?

identity-core/src/main/java/com/auth0/identity/AuthorizedIdentityProvider.java
```diff
@@ -0,0 +1,146 @@
...
+
+import com.auth0.core.Application;
+import com.auth0.identity.util.PermissionCallback;
+import com.auth0.identity.util.PermissionHandler;
+
+import java.util.List;
+
+/**
+ * Interface for allowing an IdentityProvider to request specific permissions
+ * before calling start.
```
hzalaz: Mention Android M and the possible fallbacks for older versions (it will not request permissions since they should be requested on install)

identity-core/src/main/java/com/auth0/identity/AuthorizedIdentityProvider.java
```diff
@@ -0,0 +1,146 @@
...
+ */
+public abstract class AuthorizedIdentityProvider implements IdentityProvider, PermissionCallback {
```
hzalaz: Test?

identity-core/src/main/java/com/auth0/identity/AuthorizedIdentityProvider.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /**
+     * Checks if the required permissions have been granted by the user and starts the authentication process
+     *
+     * @param activity        activity that starts the process (and will receive its response)
+     * @param serviceName     of the IdentityProvider to authenticate with
+     * @param explainIfNeeded the reason why we need the permission
+     */
+    private void checkPermissions(Activity activity, @NonNull String serviceName, boolean explainIfNeeded) {
```
hzalaz: It's ok to comment but `private` are not necessary since they are internal.

identity-core/src/main/java/com/auth0/identity/AuthorizedIdentityProvider.java
lbalmaceda: It's a helper method for the user to retry the last request:
This is to be used when a request fails because the user canceled or denied it, because he doesn't know why we are needing that android permission. The person who implements the new `AuthorizedIdentityProvider` must implement the `onPermissionRequireExplanation()` method in order to let the final user know the reason of the request. In that method, you would normally explain the reason and give the final user a button to retry the request. If this is the case, then from this method you don't have the `activity` context neither the `serviceName` to perfom a call to `start()`. This is the reason why this helper method exists.

identity-core/src/main/java/com/auth0/identity/AuthorizedIdentityProvider.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Override
+    public void start(Activity activity, @NonNull String connectionName) {
+        checkPermissions(activity, connectionName, true);
+    }
+
+    @Override
+    public void stop() {
+        identityProvider.stop();
```
hzalaz: We should clear state of this object

identity-core/src/main/java/com/auth0/identity/util/PermissionHandler.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.List;
+
+public class PermissionHandler implements ActivityCompat.OnRequestPermissionsResultCallback {
+
+    private static final int PERMISSION_REQ_CODE = 211;
+
+    private final Activity context;
+    private final AuthorizedIdentityProvider callback;
+
+    public PermissionHandler(@NonNull Activity context, @NonNull AuthorizedIdentityProvider callback) {
```
hzalaz: Why not the callback type instead of the `AuthorizedIdentityProvider`

PR #229, opened 2016-03-29
Handle Keyboard state changes

lbalmaceda:


lock/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -91,17 +98,43 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+
+    private void setupKeyboardListener(final ViewGroup container) {
+        container.getViewTreeObserver().addOnGlobalLayoutListener(new ViewTreeObserver.OnGlobalLayoutListener() {
+            @Override
+            public void onGlobalLayout() {
+                Rect r = new Rect();
+                container.getWindowVisibleDisplayFrame(r);
+                int screenHeight = container.getRootView().getHeight();
+                int keypadHeight = screenHeight - r.bottom;
+                onKeyboardStateChanged(keypadHeight > screenHeight * 0.15);
```
hzalaz: Magic number

lock/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -91,17 +98,43 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
         RelativeLayout rootView = (RelativeLayout) findViewById(R.id.com_auth0_lock_content);
         panelHolder = new ClassicPanelHolder(this, lockBus);
         rootView.addView(panelHolder, ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.MATCH_PARENT);
 
         lockBus.post(new FetchApplicationEvent());
+        setupKeyboardListener(rootContainer);
+    }
+
+    private void setupKeyboardListener(final ViewGroup container) {
+        container.getViewTreeObserver().addOnGlobalLayoutListener(new ViewTreeObserver.OnGlobalLayoutListener() {
```
hzalaz: Shouldn't we remove this when `onDestroy` is called?

lock/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -237,7 +270,9 @@ public void onFetchApplicationRequest(FetchApplicationEvent event) {
     @Subscribe
     public void onSocialAuthenticationRequest(SocialConnectionEvent event) {
         //called on social button click
-        panelHolder.showProgress(true);
+        if (!options.useBrowser()) {
```
hzalaz: pretty pointless the `if`

lock/src/main/java/com/auth0/android/lock/LockActivity.java
lbalmaceda: Yes! But there's no method to remove it prior to API 16. Anyway, making it null before `onDestroy` should prevent possible memory leaks in both cases.

PR #243, opened 2016-04-15
Use PKCE when available for /authorize calls

lbalmaceda:
Merge and deploy first https://github.com/auth0/auth0-java/pull/20

### DIFF
--- a/lock/src/main/java/com/auth0/android/lock/provider/OAuth2WebAuthProvider.java
+++ b/lock/src/main/java/com/auth0/android/lock/provider/OAuth2WebAuthProvider.java
@@ -114,11 +114,7 @@ public class OAuth2WebAuthProvider extends AuthProvider {
     @Override
     protected void requestAuth(Activity activity, String connectionName) {
         if (account.getAuthorizeUrl() == null) {
-            if (callback != null) {
-                callback.onFailure(R.string.com_auth0_lock_social_error_title, R.string.com_auth0_lock_social_invalid_authorize_url, null);
-            } else {
-                Log.w(TAG, "No callback set for web IdP authenticator");
-            }
+            callback.onFailure(R.string.com_auth0_lock_social_error_title, R.string.com_auth0_lock_social_invalid_authorize_url, null);
             return;
         }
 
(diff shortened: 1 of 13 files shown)


lock/src/main/java/com/auth0/android/lock/provider/OAuth2WebAuthProvider.java
```diff
@@ -67,21 +68,33 @@
...
     private Map<String, Object> parameters;
     private String clientInfo;
     private String lastState;
+    private PKCE pkce;
 
     public OAuth2WebAuthProvider(CallbackHelper helper, Auth0 account, @NonNull AuthCallback authCallback) {
+        this(helper, account, authCallback, false);
+    }
+
+    public OAuth2WebAuthProvider(CallbackHelper helper, Auth0 account, @NonNull AuthCallback authCallback, boolean pkce) {
```
hzalaz: Let's use a single constructor, there is no backwards compat issues here

lock/src/main/java/com/auth0/android/lock/provider/OAuth2WebAuthProvider.java
```diff
@@ -137,31 +139,44 @@ private void startAuthorization(Activity activity, Uri authorizeUri, String conn
     @Override
     public boolean authorize(Activity activity, @NonNull AuthorizeResult data) {
         if (!data.isValid(OAUTH2_REQUEST_CODE)) {
+            Log.w(TAG, "Invalid request.");
```
hzalaz: A little more info on why it failed, if you want make it debug and add the url

lock/src/main/java/com/auth0/android/lock/provider/OAuth2WebAuthProvider.java
```diff
@@ -137,31 +139,44 @@ private void startAuthorization(Activity activity, Uri authorizeUri, String conn
...
             return false;
         }
 
         final Map<String, String> values = helper.getValuesFromUri(data.getIntent().getData());
+        if (values.isEmpty()) {
+            Log.w(TAG, "Missing values from the response.");
```
hzalaz: What is `values`? remember this can be used by users of the lib so we need to provide something meaningful

lock/src/main/java/com/auth0/android/lock/provider/OAuth2WebAuthProvider.java
```diff
@@ -105,18 +114,11 @@ public void setParameters(Map<String, Object> parameters) {
...
-                callback.onFailure(R.string.com_auth0_lock_social_error_title, R.string.com_auth0_lock_social_invalid_authorize_url, null);
-            } else {
-                Log.w(TAG, "No callback set for web IdP authenticator");
-            }
+            callback.onFailure(R.string.com_auth0_lock_social_error_title, R.string.com_auth0_lock_social_invalid_authorize_url, null);
             return;
         }
 
-        //Generate random state
-        lastState = UUID.randomUUID().toString();
```
hzalaz: Not sure we need to drop this, we can use it still

lbalmaceda: I've moved it to https://github.com/auth0/Lock.Android/pull/243/files#diff-f28e3e975bb7b4fcd480dc5a7c55ecafR222

lock/src/main/java/com/auth0/android/lock/provider/OAuth2WebAuthProvider.java
lbalmaceda: At this point, I can only know if it succeeded or failed, not the cause. Maybe the logging belongs to `AuthorizeResult.isValid()`

hzalaz: Probably yes if it has more meaningful info, let's avoid adding logs without any useful info, it's just noise

PR #252, opened 2016-04-27
Add custom fields support for the sign up process

lbalmaceda:


lock/src/main/java/com/auth0/android/lock/CustomField.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    protected CustomField(Parcel in) {
+        key = in.readString();
+        hint = in.readString();
+        type = in.readInt();
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
hzalaz: If you override this you are *bound* to override `hashCode()`. Why we need to override the `equals`?

lock/src/main/java/com/auth0/android/lock/Lock.java
```diff
@@ -376,5 +372,31 @@ public Builder withProviderResolver(@NonNull AuthProviderResolver resolver) {
...
+         */
+        public Builder withSignUpFields(List<CustomField> customFields) {
+            removeDuplicatedKeys(customFields);
+            options.setCustomFields(customFields);
+            return this;
+        }
+
+        private void removeDuplicatedKeys(List<CustomField> customFields) {
+            //Remove duplicated keys
+            Map<String, CustomField> map = new LinkedHashMap<>();
```
hzalaz: I'd rather iterate the list of fields and check if I already have one with the same key (e.g. using a `Set`).

lock/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -366,13 +369,19 @@ public void onDatabaseAuthenticationRequest(DatabaseSignUpEvent event) {
...
+        }
         if (event.loginAfterSignUp()) {
             apiClient.getProfileAfter(apiClient.signUp(event.getEmail(), event.getPassword(), event.getUsername()))
-                    .addParameters(options.getAuthenticationParameters())
+                    .addParameters(authParameters)
                     .start(authCallback);
         } else {
             apiClient.createUser(event.getEmail(), event.getPassword(), event.getUsername())
-                    .addParameters(options.getAuthenticationParameters())
+                    .addParameters(authParameters)
```
hzalaz: This is actually wrong since the signup won't know how to do with them. The only parameters we need here are `'user_metadata`: {}`

lock/src/main/java/com/auth0/android/lock/views/SignUpFormView.java
```diff
@@ -121,7 +121,14 @@ public boolean validateForm() {
     @Nullable
     @Override
     public Object submitForm() {
-        return validateForm() ? getActionEvent() : null;
+        if (validateForm()) {
+            DatabaseSignUpEvent event = (DatabaseSignUpEvent) getActionEvent();
+            if (lockWidget.getConfiguration().getExtraSignUpFields().isEmpty()) {
```
hzalaz: Why not extract a method to avoid the train wreck?

PR #255, opened 2016-04-29
Change header size when entering/leaving custom fields screen

lbalmaceda:
First merge #252

image

also PR #257 centers the custom fields content.


lock/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -386,7 +392,7 @@ public void onDatabaseAuthenticationRequest(DatabaseSignUpEvent event) {
                 authParameters.put(KEY_USER_METADATA, event.extraFields());
             }
             apiClient.getProfileAfter(apiClient.signUp(event.getEmail(), event.getPassword(), event.getUsername()))
-                    .addParameters(authParameters)
+                    .addParameters(options.getAuthenticationParameters())
```
hzalaz: why we are not including user metadata?

PR #256, opened 2016-04-29
Allow to customize custom field icons

lbalmaceda:
First merge #252


hzalaz:
We should probably make the string a `StringRes` instead.

lbalmaceda:
@hzalaz both the key and the hint or just the hint?

hzalaz:
the hint, unless we are showing a the key to the user at some point

PR #260, opened 2016-05-06
Add MFA for database connections

lbalmaceda:


hzalaz:
Add screenshots and/or gifs

lbalmaceda:
screen capture on 2016-05-06 at 17-47-42

lock/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -542,7 +549,17 @@ public void onFailure(final Auth0Exception error) {
             handler.post(new Runnable() {
                 @Override
                 public void run() {
-                    String message = loginErrorBuilder.buildFrom(error).getMessage(LockActivity.this);
+                    final AuthenticationError authError = loginErrorBuilder.buildFrom(error);
+                    if (authError.getErrorType() == AuthenticationError.ErrorType.MFA_REQUIRED) {
+                        panelHolder.showProgress(false);
+                        panelHolder.showMFACodeForm(lastDatabaseLogin);
+                        return;
+                    } else if (authError.getErrorType() == AuthenticationError.ErrorType.MFA_NOT_ENROLLED) {
```
hzalaz: why `else` if previous block returns?

hzalaz: Also why this is not handled like any other error?

lbalmaceda:
I've changed a bit the MFA code screen in the last commit to like the Change Password.

Left small size / Right big size

screen shot 2016-05-10 at 1 11 13 pm

lock/src/main/java/com/auth0/android/lock/events/DatabaseLoginEvent.java
```diff
@@ -25,10 +25,13 @@
 package com.auth0.android.lock.events;
 
 
+import android.support.annotation.Nullable;
+
 public class DatabaseLoginEvent {
 
     private String usernameOrEmail;
     private String password;
+    private String MFACode;
```
hzalaz: `verificationCode`

lock/src/main/java/com/auth0/android/lock/views/ClassicPanelHolder.java
```diff
@@ -357,11 +388,18 @@ public void onKeyboardStateChanged(boolean isOpen) {
         if (customFieldsForm != null) {
             customFieldsForm.onKeyboardStateChanged(isOpen);
         }
+        if (mfaCodeFormView != null) {
```
hzalaz: Now seeing all these methods with the `if not nil then do something` for each form I think we need to handle this better since it doesn't scale and is highly error prone

lock/src/main/java/com/auth0/android/lock/views/ClassicPanelHolder.java
```diff
@@ -320,6 +348,8 @@ public void onClick(View v) {
             event = changePwdForm.submitForm();
         } else if (customFieldsForm != null) {
             event = customFieldsForm.submitForm();
+        } else if (mfaCodeFormView != null) {
```
hzalaz: Now seeing all these methods with the `if not nil then do something` for each form I think we need to handle this better since it doesn't scale and is highly error prone

lock/src/main/java/com/auth0/android/lock/views/ClassicPanelHolder.java
lbalmaceda: Do you suggest using the NullObject pattern? i.e. `Form` interface with `onKeyboardStateChanged`, `ValidForm` and `NullForm` implements it. At the beginning, all my forms are `NullForm`'s. But I always interact with a `Form` object.

lock/src/main/java/com/auth0/android/lock/views/ClassicPanelHolder.java
lbalmaceda: In this case, the first call returning an event should break and fire it. If I use NullObjects here for the forms, I'll avoid the null check of the form but I'll still need to check for a valid event.

hzalaz: It's not the null check alone is the idea if having test for every type of form we'll ever add to lock and if it's there call it's submit button, in practice it shouldn't be more than one form. Also in the last else we are not performing a null check, probably is not needed but in a 400+ lines class is not straightforward to see if it's always not null.

PR #261, opened 2016-05-09
Active MFA support

lbalmaceda:
GIF: http://g.recordit.co/8701ysCyGh.gif


hzalaz:
Use the same copies from code screen that iOS is using:

* title: https://github.com/auth0/Lock.iOS-OSX/blob/master/Lock/UI/Private/A0MFACodeViewController.m#L56
* message: https://github.com/auth0/Lock.iOS-OSX/blob/master/Lock/UI/A0MFACodeView.m#L91

simulator screen shot may 3 2016 4 07 56 pm

PR #264, opened 2016-05-12
Fullscreen state fixes

lbalmaceda:
Fix header size and contents display issues when choosing fullscreen/non-fullscreen **Lock**.
This PR also removes the dependency of the ActionBar on the `CountryCodeActivity`, and replaces it with an always present Search view. 
 
Fullscreen 
Fullscreen

Non Fullscreen
no-Fullscreen

Query
searching

EditText plate color is taken from the `Auth0.PrimaryColor` attribute of the Theme. (normally would be the same as the Submit button).


hzalaz:
@lbalmaceda rebase please

PR #265, opened 2016-05-16
Use Square Moshi instead of Jackson for Json serialization

beyondeye:
Motivation: method count: 
jackson, before proguard took more than 8300 methods. After substituting it with Moshi, method count is reduced by ~6300 methods, so that I am able to run my app that uses lock.android in debug mode without proguarding it. 
Certainly Jackson is more powerful and the orginal code was more elegant, but method counts of all the libs included with Lock.Android was a big problem for me and I think for many other potential android customers for auth0. In my personal fork of Lock.Android I also strip all code related to social authentication, since I used only passwordless, so that I can reduce method count even further.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -27,9 +27,7 @@ android {
 dependencies {
     compile fileTree(dir: 'libs', include: ['*.jar'])
     compile 'com.squareup.okhttp:okhttp:2.5.0'
-    compile 'com.fasterxml.jackson.core:jackson-databind:2.4.1'
-    compile 'com.fasterxml.jackson.core:jackson-core:2.4.1'
-    compile 'com.fasterxml.jackson.core:jackson-annotations:2.4.1'
+    compile 'com.squareup.moshi:moshi:1.1.0'
 
     testCompile libraries.testing
     testCompile 'com.squareup.okhttp:mockwebserver:2.5.0'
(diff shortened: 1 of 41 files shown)


hzalaz:
@beyondeye thanks, I am not too keen on using moshi since I haven't used it in a full fledged application yet so I'd rather be safe than sorry (even though I use a lot of stuff of Square OSS).
We are considering moving to Gson which in theory has a lower method.

BTW, which libraries are you adding that makes your method count reach the limit?

beyondeye:
Gson would be great. The problem with Gson is 
http://stackoverflow.com/questions/15485467/completing-object-construction-after-gson-deserialization 
Basically at the end you will implement by hand what Moshi adapters do. More work to do, but hey if you plan to do it would be great.
Libraries that I add: kotlin. Kotlin without reflection support is ~5100 methods. But once I add Kotlin reflection it goes up to ~16600 methods. I should probably remove it

bparent:
I'd like to second the concern about having jackson, it is a huge library. I typically like to use 3rd libs that don't introduce dependencies (like Otto, even though it is small, I tend to not use buses)

In addition, since the lib contains okhttp as well, how do you plan to support both folks who have upgraded to 3.x as well as those who 2.x still? I know its different packages, but would be great not to have to two versions of the lib.

hzalaz:
@beyondeye closing this one since soon we will release v2 (beta) and there we are using gson via auth0-java#3.0.0.

beyondeye:
nice

PR #271, opened 2016-05-23
Add Proguard rules 

lbalmaceda:


hzalaz:
Looks good but 2 things:

* Remove it from debug
* Split in several files (one per dependency), I like that better and is easier to update if we change a lib

PR #287, opened 2016-06-07
add tests for Gson usage

lbalmaceda:


lib/src/main/java/com/auth0/android/lock/utils/json/ConnectionDeserializer.java
```diff
@@ -38,7 +38,12 @@
 public class ConnectionDeserializer implements JsonDeserializer<Connection> {
     @Override
     public Connection deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws JsonParseException {
+        if (!json.isJsonObject() || json.isJsonNull()) {
```
hzalaz: If we have a utils why don't we add this there?

lib/src/main/java/com/auth0/android/lock/utils/json/JsonUtils.java
```diff
@@ -35,7 +35,7 @@
 
 import java.lang.reflect.Type;
 
-public class JsonUtils {
+public abstract class JsonUtils {
```
hzalaz: Try to make it non-public

hzalaz:
Checking serialiser code, we should improve it and specially use the context to deserialise specific attributes of valid types: e.g. String, Date, etc

hzalaz:
and rebase @lbalmaceda

lib/src/main/java/com/auth0/android/lock/utils/json/GsonDeserializer.java
```diff
@@ -24,34 +24,22 @@
...
-
-    public static void checkRequiredValue(JsonObject map, String valueName) throws JsonParseException {
+    void checkRequiredValue(JsonObject map, String valueName) throws JsonParseException {
         if (map.get(valueName) == null) {
             throw new JsonParseException("Missing value in Json: " + valueName);
         }
     }
-}
+
+    void checkValidJson(JsonElement jsonObject, Class<T> clazz) throws JsonParseException {
```
hzalaz: `checkValidJsonObject` and kill the class name since it's only used for logs

lib/src/main/java/com/auth0/android/lock/utils/json/GsonDeserializer.java
```diff
@@ -24,34 +24,22 @@
...
-        return new GsonBuilder()
-                .excludeFieldsWithoutExposeAnnotation()
-                .registerTypeAdapter(applicationType, new ApplicationDeserializer())
-                .registerTypeAdapter(strategyType, new StrategyDeserializer())
-                .registerTypeAdapter(connectionType, new ConnectionDeserializer())
-                .create();
-    }
-
-    public static void checkRequiredValue(JsonObject map, String valueName) throws JsonParseException {
+    void checkRequiredValue(JsonObject map, String valueName) throws JsonParseException {
```
hzalaz: this is better 
```java
    private <T> T requiredValue(String name, Type type, JsonObject object, JsonDeserializationContext context) throws JsonParseException {
        T value = context.deserialize(object.remove(name), type);
        if (value == null) {
            throw new JsonParseException(String.format("Missing required attribute %s", name));
        }
        return value;
    }
```

lib/src/main/java/com/auth0/android/lock/utils/json/GsonDeserializer.java
hzalaz: maybe better `assertJsonObject(JsonElement)`

PR #289, opened 2016-06-08
Handle Application without Connections

lbalmaceda:
Show a "No connections configured" message with a retry button. (Same layout is used with a different text when the application couldn't be fetched). 

http://recordit.co/ox87sWmQsF

### DIFF
--- a/lib/src/main/java/com/auth0/android/lock/views/ClassicLockView.java
+++ b/lib/src/main/java/com/auth0/android/lock/views/ClassicLockView.java
@@ -82,7 +82,7 @@ public class ClassicLockView extends LinearLayout implements View.OnClickListene
         setOrientation(VERTICAL);
         if (configuration == null) {
             Log.w(TAG, "Configuration is missing, the view won't init.");
-            showConfigurationMissingLayout();
+            showConfigurationMissingLayout(false);
         } else {
             showContentLayout();
         }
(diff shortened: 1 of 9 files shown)


lib/src/main/java/com/auth0/android/lock/views/ClassicLockView.java
```diff
@@ -154,7 +154,7 @@ private void showConfigurationMissingLayout() {
         params.gravity = Gravity.CENTER;
 
         TextView errorText = new TextView(getContext());
-        errorText.setText(R.string.com_auth0_lock_configuration_retrieving_error);
+        errorText.setText(missingConnections ? R.string.com_auth0_lock_missing_connections_message : R.string.com_auth0_lock_configuration_retrieving_error);
```
hzalaz: Why not pass the resource instead of the flag?

lib/src/main/java/com/auth0/android/lock/views/PasswordlessLockView.java
```diff
@@ -125,7 +125,7 @@ private void showConfigurationMissingLayout() {
         params.gravity = Gravity.CENTER;
 
         TextView errorText = new TextView(getContext());
-        errorText.setText(R.string.com_auth0_lock_configuration_retrieving_error);
+        errorText.setText(missingConnections ? R.string.com_auth0_lock_missing_connections_message : R.string.com_auth0_lock_configuration_retrieving_error);
```
hzalaz: Why not pass the resource instead of the flag?

PR #290, opened 2016-06-08
Refactor Custom Fields flow

lbalmaceda:
* If the custom fields are less or equal than 2, they are shown in the "first" sign up screen. If they are more, the "second" screen will show all of them.
* Consider "error" and "code" keys when parsing login errors.
* Username Style properly handled.

### DIFF
--- a/lib/src/main/java/com/auth0/android/lock/views/SignUpFormView.java
+++ b/lib/src/main/java/com/auth0/android/lock/views/SignUpFormView.java
@@ -28,24 +28,35 @@ import android.content.Context;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.util.Log;
+import android.view.Gravity;
 import android.view.KeyEvent;
 import android.view.View;
+import android.view.ViewGroup;
 import android.view.inputmethod.EditorInfo;
+import android.widget.LinearLayout;
 import android.widget.TextView;
 
 import com.auth0.android.lock.Configuration;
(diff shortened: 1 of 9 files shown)


lib/src/main/java/com/auth0/android/lock/views/SignUpFormView.java
```diff
@@ -164,10 +197,12 @@ public boolean validateForm() {
     public Object submitForm() {
         if (validateForm()) {
             DatabaseSignUpEvent event = (DatabaseSignUpEvent) getActionEvent();
-            if (!lockWidget.getConfiguration().hasExtraFields()) {
+            if (displayFewCustomFields) {
+                event.setExtraFields(getCustomFieldValues());
                 return event;
+            } else if (lockWidget.getConfiguration().hasExtraFields()) {
```
hzalaz: why else if then returns?

lib/src/main/java/com/auth0/android/lock/views/SignUpFormView.java
```diff
@@ -156,6 +183,12 @@ public boolean validateForm() {
         if (passwordInput.getVisibility() == VISIBLE) {
             valid = passwordInput.validate() && valid;
         }
+        if (displayFewCustomFields) {
+            for (int i = 0; i < fieldContainer.getChildCount(); i++) {
```
hzalaz: why not adding the flag of the if in the condition?

PR #296, opened 2016-06-13
Fix TextView lineSpacing extra padding.

lbalmaceda:
**This PR also makes bigger the Error Message**
* Fullscreen
fullscreen_long
fullscreen_short
* Normal
normal_long
normal_short


hzalaz:
I don't like the approach a lot. I'd spent a little more time trying to find a better one

lib/src/main/java/com/auth0/android/lock/views/SpacingTextView.java
```diff
@@ -0,0 +1,65 @@
...
+import android.content.Context;
+import android.os.Build;
+import android.util.AttributeSet;
+import android.widget.TextView;
+
+/**
+ * TextView that removes the extra padding added in pre Lollipop APIs when
+ * using the lineSpacingExtra attribute.
+ */
+public class SpacingTextView extends TextView {
```
hzalaz: LineSpacingTextView

PR #297, opened 2016-06-28
Password Strength Widget

lbalmaceda:
This PR adds the Password Strength Widget above the Password Field on the Sign Up form. It gets the default DB `passwordPolicy` name/level setting from the CDN.
It also removes the password length validation for the log in flow (it just checks that it's not empty) and allows white space chars in the password.

screenshot 2016-06-28 17 32 19


hzalaz:
@lbalmaceda add more margin above `Must Have` and make sure when the keyboard is displayed that the pwd error message below is visible.

lbalmaceda:
Added padding above "Must have".

Normal screen (N5x)
screenshot 2016-07-18 16 46 19

Small screen (S2)
screenshot 2016-07-18 16 49 55

PR #299, opened 2016-07-13
missing "{" on line 103

zeevgl:
missing "{" on line 103


hzalaz:
Thanks, we'll fix it ourselves since it might be easier than rebasing your PR

PR #310, opened 2016-07-27
Make PKCE enabled by default

lbalmaceda:


hzalaz:
Let's rename the builder option to `usingImplicitGrant()`

PR #312, opened 2016-07-29
Send Accept-Language Header (v1)

lbalmaceda:
Defaults to language "en" if the default language wasn't available/found.


core/src/main/java/com/auth0/api/internal/RequestFactory.java
```diff
@@ -92,4 +116,10 @@ public static void setClientInfo(String clientInfo) {
         }
         return request;
     }
+
+    private static <T> ParameterizableRequest<T> addLocaleHeader(ParameterizableRequest<T> request) {
+        String language = Locale.getDefault().toString();
+        request.addHeader("Accept-Language", language != null ? language : "en");
```
nikolaseu: check for empty string, not null

PR #318, opened 2016-08-05
Allow to cancel requests. (v1)

lbalmaceda:
This PR fixes some NPE occurring when a response arrived and the activity was in the background.


nikolaseu:
LGTM

PR #319, opened 2016-08-08
Request the user to accept Terms&Policy before Sign Up

lbalmaceda:
This affects database connections only. The dev must set the option on the builder first:

```java
Lock lock = new Lock.Builder(account)
// ...
.setMustAcceptTerms(true)
.build();
```

When he completes the fields with valid data and press the submit button, a dialog will prompt asking for acceptance.

### DIFF
--- a/lib/src/test/java/com/auth0/android/lock/OptionsTest.java
+++ b/lib/src/test/java/com/auth0/android/lock/OptionsTest.java
@@ -11,7 +11,9 @@ import com.auth0.android.lock.enums.SocialButtonStyle;
 import com.auth0.android.lock.enums.UsernameStyle;
 
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
 import org.robolectric.RobolectricGradleTestRunner;
 import org.robolectric.annotation.Config;
@@ -44,6 +46,9 @@ public class OptionsTest {
     private static final String SCOPE_OPENID_OFFLINE_ACCESS = "openid offline_access";
 
     private Options options;
+
+    @Rule
+    public ExpectedException exception = ExpectedException.none();
 
     @Before
     public void setUp() throws Exception {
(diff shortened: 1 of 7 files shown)


lib/src/test/java/com/auth0/android/lock/OptionsTest.java
```diff
@@ -94,23 +107,20 @@ public void shouldNotSetPrivacyPolicyURLWhenInvalidURL() {
...
+        assertThat(options.getTermsURL(), is(notNullValue()));
         assertThat(options.getTermsURL(), is(equalTo(parceledOptions.getTermsURL())));
     }
 
     @Test
-    public void shouldNotSetTermsOfServiceURLWhenInvalidURL() {
-        Options options = new Options();
-        options.setAccount(auth0);
+    public void shouldNotSetTermsOfServiceURLWhenInvalidURL() throws Exception {
         options.setTermsURL("an-invalid/url");
```
nikolaseu: slightly OT but shouldn't this throw IAE instead of leaving a null value ?

lib/src/test/java/com/auth0/android/lock/OptionsTest.java
```diff
@@ -94,23 +107,20 @@ public void shouldNotSetPrivacyPolicyURLWhenInvalidURL() {
...
-        options.setAccount(auth0);
+    public void shouldSetTermsOfServiceURL() throws Exception {
         options.setTermsURL("https://valid.url/terms");
 
         Parcel parcel = Parcel.obtain();
         options.writeToParcel(parcel, 0);
         parcel.setDataPosition(0);
 
         Options parceledOptions = Options.CREATOR.createFromParcel(parcel);
+        assertThat(options.getTermsURL(), is(notNullValue()));
```
nikolaseu: what if `options.getTermsURL()` return "some other url that I'm mixing with privacy url for example" ?

lib/src/test/java/com/auth0/android/lock/OptionsTest.java
lbalmaceda: If the given url it's not valid, the it will default to Auth0's terms/policy urls. I was trying to avoid throwing an exception at this point, and just warning the user that the default will be used on the LogCat. Anyway, I can add that change.

Issue #225, opened 2016-03-21
Add support for custom social OAuth2 connections


plusCubed:
https://auth0.com/docs/connections/social/oauth2
That was the manual way to add custom OAuth2 connections - recently the Custom Social Connections extension was added to the Auth0 dashboard too.

Right now, Lock shows the custom connection as "Login with OAuth2" (with a generic logo and color), and opens the wrong authorization URL, `connection=oauth2` instead of `connection=[whatever the actual custom connection name is]`.

mdumrauf:
Needing this too

PR #320, opened 2016-08-08
Use first available connection name when authenticating with OAuth

lbalmaceda:
OAuth authentication was sending the Strategy name as the connection name parameter. This PR changes that to the first Connection name that it founds inside the Strategy, or the Strategy name if no connections were available for that Strategy.

### DIFF
--- a/lib/src/main/java/com/auth0/android/lock/utils/json/Strategy.java
+++ b/lib/src/main/java/com/auth0/android/lock/utils/json/Strategy.java
@@ -25,7 +25,7 @@
 package com.auth0.android.lock.utils.json;
 
 
-import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 
 import com.auth0.android.lock.utils.Strategies;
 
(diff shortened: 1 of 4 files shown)


nikolaseu:
👍

lib/src/main/java/com/auth0/android/lock/utils/json/Strategy.java
```diff
@@ -71,4 +73,15 @@ public boolean isActiveFlowEnabled() {
...
+
+    /**
+     * Returns the name of the first connection found in this strategy. When no connections available,
+     * it will default to the strategy name.
+     *
+     * @return the first connection found or the strategy name if no connections are available.
+     */
+    @NonNull
+    public String getDefaultConnectionName() {
+        return !connections.isEmpty() ? connections.get(0).getName() : this.name;
```
hzalaz: This is not quite right for all strategies since returning the strategy name only applies to social ones. We should return `null` or avoid using a strategy without connections

lbalmaceda: We can improve the `Configuration` class by removing non social Strategies without connection in the filter process, or even considering social Strategies too. Maybe logging something to let the user know about that.

lbalmaceda:
Related to https://github.com/auth0/Lock.Android/issues/225

PR #325, opened 2016-08-16
Custom Style for Social Buttons

lbalmaceda:
With this changes the user can style each `Connection` individually. If a Connection it's not customized by the user, it will default to the Lock Strategy's defaults. i.e. "facebook" connection will default to the facebook logo, color and title.

### Usage:
Create a new style with the following keys:
```xml
<style name="MyConnectionStyle" parent="Lock.Theme.AuthStyle">
        <item name="Auth0.BackgroundColor">@color/my_connection_color</item>
        <item name="Auth0.Name">@string/my_connection_name</item>
        <item name="Auth0.Logo">@drawable/my_connection_logo</item>
</style>
```

Then in the `Lock.Builder` instance set the style you'd like to override: `builder.withAuthStyle("my-connection-name", R.style.MyConnectionStyle);`

When Lock needs to show a button for that Connection, it will get the Style you defined. This means that for each Connection inside a Strategy, it will create a new Style.


lib/src/main/java/com/auth0/android/lock/views/AuthConfig.java
```diff
@@ -0,0 +1,196 @@
...
+
+    /**
+     * Retrieves the resource id of the given Style index.
+     *
+     * @param context a valid Context
+     * @param index   The index to search on the Style definition.
+     * @return the id if found or -1.
+     */
+    int getIdForResource(@NonNull Context context, @StyleableRes int index) {
+        final int[] attrs = new int[]{index};
```
nikolaseu: Maybe it's better to use 
```TypedArray obtainStyledAttributes (AttributeSet set, int[] attrs, int defStyleAttr, int defStyleRes)```
at the constructor, using the default Theme as param, like its done in the custom views. 
Would be less code and also the resources would only be looked up once. We'll need to add a context to the constructor.

File: lib/src/main/java/com/auth0/android/lock/Configuration.java [hunk 2]
```diff
@@ -322,6 +327,14 @@ private void parseLocalOptions(Options options) {
         this.privacyURL = options.getPrivacyURL() == null ? "https://auth0.com/privacy" : options.getPrivacyURL();
     }
 
+    @StyleRes
+    public int authStyleForConnection(String strategy, String connection) {
+        if (authStyles.containsKey(connection)) {
+            return authStyles.get(connection);
+        }
+        return AuthConfig.styleForStrategy(strategy);
```
nikolaseu: `public static` is hard to test since you can't mock/inject

lib/src/main/java/com/auth0/android/lock/views/SocialView.java
```diff
@@ -75,6 +76,15 @@ private void init(boolean smallButtons) {
         setLayoutParams(new ViewGroup.LayoutParams(ViewGroup.LayoutParams.WRAP_CONTENT, ViewGroup.LayoutParams.WRAP_CONTENT));
     }
 
+    private List<AuthConfig> generateAuthConfigs(List<Strategy> socialStrategies) {
+        List<AuthConfig> configs = new ArrayList<>();
+        for (Strategy s : socialStrategies) {
+            int style = lockWidget.getConfiguration().authStyleForConnection(s.getName(), s.getDefaultConnectionName());
+            configs.add(new AuthConfig(s.getName(), s.getDefaultConnectionName(), style));
```
nikolaseu: `authStyleForConnection` uses AuthConfig for default values, and here we create again. Why don't just have the array of `AuthConfig` on `Configuration`? And the `AuthConfig` could have a constructor without custom style that would use the default (so the static method can be private)

lib/src/main/java/com/auth0/android/lock/Configuration.java
lbalmaceda: The tests run with Robolectric. A resource folder is provided with the `styles.xml` included in the original lib. The test I wrote (`AuthConfigTest`) checks that given a `Strategy` name I get the correct Style ID ref.

nikolaseu: In this case there's no need to expose a public static method, just return -1 or something like that when there's no config.
So the defaults are handled in AuthConfig itself, when constructed without a StyleRes (-1). Then the static method is private and it doesn't even need to be static

nikolaseu: about my comment `public static is hard to test since you can't mock/inject`
I was trying to note that with the static method there, the test for this method would have to repeat all the tests for the static method itself because you can't check that you're calling a static method

PR #329, opened 2016-08-17
Force init lock [Breaking Change]

lbalmaceda:
* Remove `Lock.onCreate` and `PasswordlessLock.onCreate` methods. Now you initialize the instance in the `Lock.Builder.build(Activity)` method. [Breaking Change]
* Update the README to reflect the `onCreate` removal.
* Add new builder initializer method `public static Builder newBuilder(@NonNull LockCallback callback)` passing only the callback. The `Auth0` account will be generated using the strings defined in the `strings.xml` resource file (or any other file inside the android resources). If the accounts details can't be found, an exception will be thrown.

### DIFF
--- a/lib/src/main/java/com/auth0/android/lock/Lock.java
+++ b/lib/src/main/java/com/auth0/android/lock/Lock.java
@@ -85,7 +85,6 @@ public class Lock {
 
     /**
      * Creates a new Lock.Builder instance with the given account and callback.
-     * ¬
      *
      * @param account  details to use against the Auth0 Authentication API.
      * @param callback that will receive the authentication results.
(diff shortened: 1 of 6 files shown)


lib/src/main/java/com/auth0/android/lock/PasswordlessLock.java
```diff
@@ -190,19 +193,32 @@ public Builder(Auth0 account, LockCallback callback) {
...
+                }
             }
             if (callback == null) {
                 Log.e(TAG, "You need to specify the callback object to receive the Authentication result.");
                 throw new IllegalStateException("Missing callback.");
             }
             Log.v(TAG, "PasswordlessLock instance created");
-            return new PasswordlessLock(options, callback);
+
+            if (options.getAccount().getTelemetry() != null) {
```
nikolaseu: shouldn't this be `== null` ?

lib/src/main/java/com/auth0/android/lock/PasswordlessLock.java
```diff
@@ -190,19 +193,32 @@ public Builder(Auth0 account, LockCallback callback) {
...
+                Log.v(TAG, String.format("Using Telemetry %s (%s) and Library %s", Constants.LIBRARY_NAME, com.auth0.android.lock.BuildConfig.VERSION_NAME, BuildConfig.VERSION_NAME));
+                options.getAccount().setTelemetry(new Telemetry(Constants.LIBRARY_NAME, com.auth0.android.lock.BuildConfig.VERSION_NAME, BuildConfig.VERSION_NAME));
```
nikolaseu: if telemetry is required, maybe this can be handled in Auth0 itself? on `getTelemetry()` check if it's null and then create and save the default

lib/src/main/java/com/auth0/android/lock/Lock.java
```diff
@@ -201,19 +203,32 @@ public Builder(Auth0 account, LockCallback callback) {
...
+                }
             }
             if (callback == null) {
                 Log.e(TAG, "You need to specify the callback object to receive the Authentication result.");
                 throw new IllegalStateException("Missing callback.");
             }
             Log.v(TAG, "Lock instance created");
-            return new Lock(options, callback);
+
+            if (options.getAccount().getTelemetry() != null) {
```
nikolaseu: same, `==`

PR #330, opened 2016-08-18
Check that requested tokens are present on the result.

lbalmaceda:
* Remove localized error message when Authentication fails. This was used when calling `AuthenticationCallback` with a `LockException`.
* Check for requested scopes, and validate that the `Credentials` received contains them.
* Fail with `callback.onError(LockException)` when a requested token is missing from the response, even if the response was successful.


lib/src/main/java/com/auth0/android/lock/Lock.java
```diff
@@ -162,7 +163,11 @@ private void processEvent(Intent data) {
         switch (action) {
             case Constants.AUTHENTICATION_ACTION:
                 Log.v(TAG, "AUTHENTICATION action received in our BroadcastReceiver");
-                callback.onEvent(LockEvent.AUTHENTICATION, data);
+                if (data.getExtras().containsKey("error")) {
```
nikolaseu: `Constants.ERROR_EXTRA` ?

PR #333, opened 2016-08-19
Refactor AuthProviderResolver [Breaking Change]

lbalmaceda:
Previously, the user needed to define it's own `AuthProviderResolver` when he was notified of authentication events, asking for an `AuthProvider` given a `connectionName`. To simplify that, now the user calls the builder passing handlers:

```java
builder. withAuthHandlers(Arrays.asList(new MyHandler(), new FacebookHandler()));
```
Each handler can decide whether it can or not handle a given strategy/connection name. Because it's a List, the order matters and the first match will return an `AuthProvider`. If no provider is found, Lock will default to the `WebAuthProvider`.

This way we can also provide our own handlers in the social native libs.


lib/src/main/java/com/auth0/android/lock/Lock.java
```diff
@@ -374,13 +374,13 @@ public Builder loginAfterSignUp(boolean login) {
...
-         * Uses the given AuthProviderResolver to ask for Native IdentityProviders.
+         * Uses the given AuthHandlers to query for AuthProviders on a new authentication request.
          *
-         * @param resolver the AuthProviderResolver to use
+         * @param handlers that Lock will query for AuthProviders.
          * @return the current builder instance
          */
-        public Builder withProviderResolver(@NonNull AuthProviderResolver resolver) {
-            ProviderResolverManager.set(resolver);
+        public Builder withAuthHandlers(@NonNull List<AuthHandler> handlers) {
```
hzalaz: Variable args

lib/src/test/java/com/auth0/android/lock/provider/AuthResolverTest.java
```diff
@@ -24,34 +24,39 @@
...
-        AuthProviderResolver resolver = ProviderResolverManager.get();
+        AuthProvider provider = Mockito.mock(AuthProvider.class);
+        AuthHandler handler = Mockito.mock(AuthHandler.class);
+        Mockito.when(handler.providerFor("a", "a")).thenReturn(provider);
+        AuthResolver.setAuthHandlers(Arrays.asList(handler));
 
-        assertThat(resolver, notNullValue());
-        assertThat(resolver, equalTo(customResolver));
+        assertThat(AuthResolver.providerFor("", ""), nullValue());
+        assertThat(AuthResolver.providerFor("a", "a"), is(equalTo(provider)));
```
nikolaseu: since the param is a list, maybe test at least two providers? "a" and "b" ?

lib/src/main/java/com/auth0/android/lock/provider/AuthResolver.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        AuthProvider provider = null;
+        for (AuthHandler p : authHandlers) {
+            provider = p.providerFor(strategy, connection);
+            if (provider != null) {
+                break;
+            }
+        }
+        return provider;
```
nikolaseu: maybe it's simpler this way?
```
         for (AuthHandler p : authHandlers) {
             final AuthProvider provider = p.providerFor(strategy, connection);
             if (provider != null) {
                 return provider;
             }
         }
         return null;
```

lib/src/main/java/com/auth0/android/lock/views/SocialView.java
```diff
@@ -86,8 +87,8 @@ private void init(boolean smallButtons) {
     }
 
     @Override
-    public void onConnectionClicked(String connectionName) {
-        lockWidget.onSocialLogin(new SocialConnectionEvent(connectionName));
+    public void onAuthenticationRequest(@NonNull String strategy,@NonNull String connection) {
+        lockWidget.onSocialLogin(new SocialConnectionEvent(strategy, connection));
```
nikolaseu: this is just a comment, but maybe the name of this method should better reflect what it does? now it's clear that an authentication has been requested, but the `onSocialLogin` is not that clear (`onSocialLoginRequest`?).

lib/src/main/java/com/auth0/android/lock/provider/AuthResolver.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.List;
+
+
+/**
+ * Holds instances of AuthHandlers that can be used to query for AuthProviders given
+ * a strategy and connection name.
+ * When no AuthProvider is matched, it will return null
+ */
+public abstract class AuthResolver {
```
nikolaseu: and `final`, if possible

lbalmaceda: @nikolaseu Both options are good. Although using a private constructor has the advantage of not letting the user to extend the abstract class and instantiate objects. I will change it. 👍

lib/src/main/java/com/auth0/android/lock/provider/AuthResolver.java
lbalmaceda: I think both are acceptable and clear.. But the one I did has 1 return point at the end, maybe that helps a bit reading the code?

lib/src/main/java/com/auth0/android/lock/views/SocialView.java
lbalmaceda: I agree. 👍

hzalaz: reactions please 😄

Issue #332, opened 2016-08-19
Lock SSO Username Fails Validation


codajustin:
The Regular Expression for Username validation is too restrictive - it is only allowing a max of 15 alpha-numeric characters and underscores (_).  This automatically excludes schemes that use firstname.lastname like Office365 or Microsoft Azure AD for example.

Attempted to build Lock with .withUsernameStyle(UsernameStyle.EMAIL) but when it hits SSO, it automatically changes to Username.

The current RegEx in ValidatedInputView:
`public static final String USERNAME_REGEX = "^[a-zA-Z0-9_]{1,15}$";`

![sso](https://cloud.githubusercontent.com/assets/21108254/17790595/d085e4f2-655c-11e6-931c-c7323d737650.png)

Using Lock-2.0.0-beta3. 
Tested on Android M: Samsung Galaxy Note 5

PR #334, opened 2016-08-19
Change username validation for SSO connections

lbalmaceda:
This PR fixes https://github.com/auth0/Lock.Android/issues/332.


lib/src/main/java/com/auth0/android/lock/views/ValidatedInputView.java
```diff
@@ -95,6 +96,7 @@
         int DATE = 7;
         int TEXT_NAME = 8;
         int NUMBER = 9;
+        int SSO_USERNAME = 10;
```
hzalaz: I am not buying this new name, its kind of cryptic. Ideally it's just a non-empty validation why we need to have a `SSO_USERNAME`? The style should be the same as the username one, and it's still a username in a sense

PR #335, opened 2016-08-23
Flatten Strategies into Connections

lbalmaceda:
Merge Strategy & Connection into a single class --> `Connection`. This flattens the POJO structure for better handling inside lock.


lib/src/main/java/com/auth0/android/lock/Configuration.java
```diff
@@ -121,144 +112,82 @@ public Connection getDefaultDatabaseConnection() {
...
-    }
-
-    public Strategy getActiveDirectoryStrategy() {
-        return activeDirectoryStrategy;
-    }
-
     @Nullable
-    public Strategy getDefaultPasswordlessStrategy() {
-        if (passwordlessStrategies.isEmpty()) {
+    public Connection getDefaultPasswordlessConnection() {
```
hzalaz: `getPasswordlessConnection`

lib/src/main/java/com/auth0/android/lock/Configuration.java
```diff
@@ -97,19 +90,17 @@ public Configuration(Application application, Options options) {
...
+        this.defaultDatabaseConnection = filterDatabaseConnections(application.getConnections(), connectionSet, defaultDatabaseName);
+        this.enterpriseConnections = filterConnections(application.getConnections(), connectionSet, AuthType.ENTERPRISE);
+        this.passwordlessConnections = filterConnections(application.getConnections(), connectionSet, AuthType.PASSWORDLESS);
+        this.socialConnections = filterConnections(application.getConnections(), connectionSet, AuthType.SOCIAL);
         this.application = application;
         parseLocalOptions(options);
         boolean atLeastOneSocialModeEnabled = allowLogIn || allowSignUp;
         boolean atLeastOneDatabaseModeEnabled = atLeastOneSocialModeEnabled || allowForgotPassword;
-        this.classicLockAvailable = (!socialStrategies.isEmpty() && atLeastOneSocialModeEnabled) || (!enterpriseStrategies.isEmpty() && allowLogIn)
+        this.classicLockAvailable = (!socialConnections.isEmpty() && atLeastOneSocialModeEnabled) || (!enterpriseConnections.isEmpty() && allowLogIn)
```
hzalaz: Do we need this?

lib/src/main/java/com/auth0/android/lock/utils/EnterpriseConnectionMatcher.java
```diff
@@ -44,20 +44,20 @@
...
+        this.connections = new ArrayList<>();
+        if (connections == null) {
             return;
         }
 
-        for (Strategy s : strategies) {
-            if (s.getType() == Strategies.Type.ENTERPRISE) {
-                this.strategies.add(s);
+        for (Connection s : connections) {
+            if (s.getType() == AuthType.ENTERPRISE) {
```
hzalaz: Why we need to do this here? are these connections supposed to be HRD enabled connections?

lib/src/main/java/com/auth0/android/lock/utils/json/Application.java
```diff
@@ -46,17 +39,10 @@
...
     private String subscription;
     private boolean hasAllowedOrigins;
-    private List<Strategy> strategies;
-    private List<Strategy> socialStrategies;
-    private List<Strategy> enterpriseStrategies;
-    private List<Strategy> passwordlessStrategies;
-    private Strategy databaseStrategy;
+    private List<Connection> connections;
 
     public Application() {
```
hzalaz: Do we still need this class? Configuration has this info too?

hzalaz:
We should move all these client info related classes to an `internal` package (since we can't change the modifier access)

lib/src/main/java/com/auth0/android/lock/Configuration.java
lbalmaceda: @hzalaz This values are queried before showing the Lock screen to determine not only if the connections are available (after parsing the response from the cdn) but to also check if the user has disabled some screens. i.e.: 
* If `social connections` are available but both `sign up / log in screen` are disabled, it won't show it.
* If `enterprise connections` are available but `log in screen` is disabled, it won't show it.

I can do this same boolean logic on the view side if that's what you want.

lib/src/main/java/com/auth0/android/lock/utils/EnterpriseConnectionMatcher.java
lbalmaceda: @hzalaz It was a safe-filter to only use enterprise connections in this class. But the connections come from `configuration.getEnterpriseConnections()` so this can be removed. Also, the class can be renamed to something more generic like `ConnectionDomainMatcher`.

lib/src/main/java/com/auth0/android/lock/Lock.java
```diff
@@ -206,6 +209,16 @@ public Lock build(@NonNull Activity activity) {
                 Log.e(TAG, "You need to specify the callback object to receive the Authentication result.");
                 throw new IllegalStateException("Missing callback.");
             }
+            if (!options.allowForgotPassword() && !options.allowLogIn() && !options.allowSignUp()) {
```
hzalaz: the if-else are not needed

lib/src/main/java/com/auth0/android/lock/enums/Strategies.java
```diff
@@ -0,0 +1,128 @@
+package com.auth0.android.lock.enums;
```
hzalaz: let's remove the `enums` package for something more meaningful

lib/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
```diff
@@ -400,14 +404,14 @@ public void onCountryCodeChangeRequest(CountryCodeChangeEvent event) {
...
-        if (configuration.getDefaultPasswordlessStrategy() == null) {
+        if (configuration.getPasswordlessConnection() == null) {
             Log.w(TAG, "There is no default Passwordless strategy to authenticate with");
             return;
         }
 
         lockView.showProgress(true);
         AuthenticationAPIClient apiClient = new AuthenticationAPIClient(options.getAccount());
-        String connectionName = configuration.getFirstConnectionOfStrategy(configuration.getDefaultPasswordlessStrategy());
+        String connectionName = configuration.getPasswordlessConnection().getName();
```
hzalaz: Should we handle the case where the connection is null?

lib/src/main/java/com/auth0/android/lock/enums/Strategies.java
```diff
@@ -0,0 +1,128 @@
...
+import static com.auth0.android.lock.enums.Strategies.Yahoo;
+import static com.auth0.android.lock.enums.Strategies.Yammer;
+import static com.auth0.android.lock.enums.Strategies.Yandex;
+
+@StringDef({Auth0, Email, SMS, Amazon, AOL, Baidu, BitBucket, Box, Dropbox, Dwolla, EBay, Evernote, EvernoteSandbox, Exact, Facebook,
+        Fitbit, Github, GooglePlus, Instagram, Linkedin, Miicard, Paypal, PlanningCenter, RenRen, Salesforce, SalesforceSandbox, Shopify,
+        Soundcloud, TheCity, TheCitySandbox, ThirtySevenSignals, Twitter, VK, Weibo, WindowsLive, Wordpress, Yahoo, Yammer, Yandex, UnknownSocial,
+        ActiveDirectory, ADFS, Auth0LDAP, Custom, GoogleApps, GoogleOpenId, IP, MSCRM, Office365, PingFederate, SAMLP, Sharepoint, Waad})
+@Retention(RetentionPolicy.SOURCE)
+public @interface Strategies {
```
hzalaz: The need this because?

lib/src/main/java/com/auth0/android/lock/internal/json/Strategy.java
```diff
@@ -0,0 +1,33 @@
+package com.auth0.android.lock.internal.json;
+
+import android.support.annotation.NonNull;
+
+import java.util.List;
+
+import static com.auth0.android.util.CheckHelper.checkArgument;
+
+class Strategy {
```
hzalaz: Can't we kill this now, serves little purpose?

lib/src/main/java/com/auth0/android/lock/enums/Strategies.java
lbalmaceda: @hzalaz maybe `constants` or `annotations` ?

lib/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
lbalmaceda: @hzalaz That's handled before showing the widget, it will show an error if no connections are available for the "Passwordless mode" (needs social || passwordless) -> https://github.com/auth0/Lock.Android/blob/1d31ce2fcd3a46cea6728070f3158317f2976837/lib/src/main/java/com/auth0/android/lock/views/PasswordlessLockView.java#L115
Also, in case the event is received but again no passwordless connection is available, it will break before calling authenticate -> https://github.com/auth0/Lock.Android/pull/335/files#diff-396f039685bb2421bd638efb3baadd3eR407

lib/src/main/java/com/auth0/android/lock/enums/Strategies.java
lbalmaceda: @hzalaz Not needed. Will be deleted. 👍

lib/src/main/java/com/auth0/android/lock/internal/json/Strategy.java
lbalmaceda: @hzalaz It was only used on the JSON parsing stage for better readability. Now I've moved all the logic in the `ApplicationDeserializer` class.

lib/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
hzalaz: Ok, if there is a test to avoid a subtle change in order of execution to break the assumption

PR #340, opened 2016-09-06
Filter social strategies by connection name

lbalmaceda:
Filter social connections comparing their name instead of the strategy name. Ignore those strategies without connections.

### DIFF
--- a/lib/src/main/java/com/auth0/android/lock/Configuration.java
+++ b/lib/src/main/java/com/auth0/android/lock/Configuration.java
@@ -224,12 +224,13 @@ public class Configuration {
     }
 
     private List<Strategy> filterSocialStrategies(List<Strategy> strategies, Set<String> connections) {
-        if (strategies == null || connections.isEmpty()) {
-            return strategies;
+        if (strategies == null) {
+            return null;
         }
         List<Strategy> filtered = new ArrayList<>(strategies.size());
         for (Strategy strategy : strategies) {
-            if (connections.contains(strategy.getName())) {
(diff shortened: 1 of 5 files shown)


lib/src/main/java/com/auth0/android/lock/utils/json/Strategy.java
```diff
@@ -86,6 +86,6 @@ public String getDefaultConnectionName() {
         if (!connections.isEmpty()) {
             return connections.get(0).getName();
         }
-        return strategyMetadata.getType() == Strategies.Type.SOCIAL ? this.name : null;
+        return null;
```
hzalaz: I think this can be solved in 1 line

lib/src/main/java/com/auth0/android/lock/Configuration.java
```diff
@@ -224,12 +224,13 @@ private Strategy filterStrategy(Strategy strategy, Set<String> connections) {
...
     private List<Strategy> filterSocialStrategies(List<Strategy> strategies, Set<String> connections) {
-        if (strategies == null || connections.isEmpty()) {
-            return strategies;
+        if (strategies == null) {
+            return null;
         }
         List<Strategy> filtered = new ArrayList<>(strategies.size());
         for (Strategy strategy : strategies) {
-            if (connections.contains(strategy.getName())) {
+            final boolean allowed = connections.isEmpty() || connections.contains(strategy.getDefaultConnectionName());
```
hzalaz: Should we check all connections not just one?

PR #341, opened 2016-09-06
Show button when only one enterprise connection is available

lbalmaceda:
* 1 or more enterprise connection with Social connections (will try to match domain)
device-2016-09-07-140806

* Single enterprise connection:
device-2016-09-06-180727


hzalaz:
rebase @lbalmaceda

PR #345, opened 2016-09-15
Fetch username min/max length validation from the CDN

lbalmaceda:
If either min/max values are missing, it will default to `min=1` and `max=15`. This PR also separates Connection into interfaces: 

* `DatabaseConnection` with getters for common booleans, and the min/max username length.
* `OAuthConnection` with getters for the domains set and the "/ro enabled".
* `PasswordlessConnection`


lib/src/main/java/com/auth0/android/lock/internal/configuration/ApplicationFetcher.java
```diff
@@ -71,13 +70,13 @@ public ApplicationFetcher(@NonNull Auth0 account, @NonNull OkHttpClient client)
...
-    public void fetch(@NonNull AuthenticationCallback<List<Connection>> callback) {
+    public void fetch(@NonNull AuthenticationCallback<Configuration> callback) {
         makeApplicationRequest(callback);
     }
 
-    private void makeApplicationRequest(final AuthenticationCallback<List<Connection>> callback) {
-        Uri uri = Uri.parse(account.getConfigurationUrl()).buildUpon().appendPath("client")
-                .appendPath(account.getClientId() + ".js").build();
+    private void makeApplicationRequest(final AuthenticationCallback<Configuration> callback) {
+        Uri uri = Uri.parse(options.getAccount().getConfigurationUrl()).buildUpon().appendPath("client")
```
hzalaz: We should pass along the Account only not the whole options object

lib/src/main/java/com/auth0/android/lock/internal/configuration/ApplicationFetcher.java
```diff
@@ -103,7 +102,7 @@ public void onResponse(Response response) {
                 }
 
                 Log.i(TAG, "Application received!");
-                callback.onSuccess(connections);
+                callback.onSuccess(new Configuration(connections, options));
```
hzalaz: It's better to leave the responsibility of building the configuration to the caller

lib/src/main/java/com/auth0/android/lock/internal/configuration/Connection.java
```diff
@@ -0,0 +1,186 @@
...
+            case "waad":
+                return AuthType.ENTERPRISE;
+            default:
+                return AuthType.SOCIAL;
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    public <T> T getValueForKey(String key) {
+        return (T) this.values.get(key);
```
hzalaz: This might cause `ClassCastException` we need to check if it can be casted first

lib/src/main/java/com/auth0/android/lock/internal/configuration/Connection.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    @Override
+    public boolean booleanForKey(String key) {
+        Boolean value = getValueForKey(key);
+        if (value == null) {
```
hzalaz: This should work
```java
return value != null && value
```

lib/src/main/java/com/auth0/android/lock/internal/configuration/Connection.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    /**
+     * Creates a new Connection given a Strategy name and the map of values.
+     *
+     * @param strategy strategy name for this connection
+     * @param values   additional values associated to this connection
+     * @return a new instance of Connection. Can be either DatabaseConnection, PasswordlessConnection or OAuthConnection.
+     */
+    public static Connection connectionFor(@NonNull String strategy, Map<String, Object> values) {
```
hzalaz:
```java
public static Connection newConnectionFor(@NonNull String strategy, Map<String, Object> values)
```

lib/src/main/java/com/auth0/android/lock/internal/configuration/Connection.java
```diff
@@ -0,0 +1,186 @@
...
+        if (minUsernameLength < MIN_USERNAME_LENGTH || minUsernameLength > maxUsernameLength) {
+            minUsernameLength = MIN_USERNAME_LENGTH;
+            maxUsernameLength = MAX_USERNAME_LENGTH;
+        }
+    }
+
+    private int intValue(@Nullable Object object) {
+        int value = 0;
+        try {
+            value = object == null ? 0 : (int) Double.parseDouble(String.valueOf(object));
```
hzalaz: Values will be Integer not Double

lib/src/main/java/com/auth0/android/lock/internal/configuration/BaseConnection.java
```diff
@@ -0,0 +1,35 @@
...
+    String getStrategy();
+
+    /**
+     * Returns a value using its key
+     *
+     * @param key a key
+     * @param <T> type of value to return
+     * @return a value
+     */
+    <T> T getValueForKey(String key);
```
hzalaz:
```java
<T> T valueForKey(String key)
```

or make `booleanForKey` be a getter too

lib/src/main/java/com/auth0/android/lock/internal/configuration/Connection.java
lbalmaceda: @hzalaz Yes. That case was tested in the `booleanForKey` [method](https://github.com/auth0/Lock.Android/blob/e771daf22fb16272016b02b9e4adf4c20996a791/lib/src/test/java/com/auth0/android/lock/internal/configuration/ConnectionTest.java#L120-L126). Without adding parameters, we can try-catch the `ClassCastException` and return a `null` value. The other option is to also receive the `T` class in the parameters and ask if the value of that key is an instance of `T` class.

hzalaz: Catching `ClassCastException` is never a good idea so we need to provide the Type we expect

lib/src/main/java/com/auth0/android/lock/internal/configuration/Connection.java
lbalmaceda: We can't fix the structure of that json dictionary as it may change in the future, so we need to set the target to a generic map. When deserializing a `Map<String, Object>` *Gson* will try to convert "string numbers" to double so not to lose any precision. That's why we need to get the double first and then remove that precision. More info [here](http://stackoverflow.com/questions/15507997/how-to-prevent-gson-from-expressing-integers-as-floats).

nikolaseu: Gson already does that for you, so `object` should already be an instance of `Double`, or a `String` (if the server sends it between `"`)

hzalaz: The server will **always** send an integer field as a number or string. If it sends 1.0 it's invalid so we should not consider it a valid one

PR #353, opened 2016-10-01
Improve dependencies versioning (externalized versions)

segunfamisa:
This Pull Request helps to improve the dependency versioning in the library.

Externalize dependency versions by using gradle extra properties, instead of hardcoding and probably duplicating version numbers.

It also makes version upgrades easy.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,13 +1,13 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 24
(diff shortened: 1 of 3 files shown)


hzalaz:
Thanks, but we'd rather stick without externalized versions in another file since the lib only has 1 gradle file, the other is the application which will only break for us. The main reason is to allow AS to notify us of outdated google deps and plugins, I can do that for the dependencies but it's trickier for plugins and build tools

PR #360, opened 2016-10-12
Non-empty username validation for custom/imported connections

lbalmaceda:
Validate that the username it's non-empty, regardless it's length and format, if the connection to be used it's a custom database connection or an imported one.

### DIFF
--- a/lib/src/main/java/com/auth0/android/lock/internal/configuration/Connection.java
+++ b/lib/src/main/java/com/auth0/android/lock/internal/configuration/Connection.java
@@ -158,9 +158,14 @@ public class Connection implements BaseConnection, DatabaseConnection, OAuthConn
 
     private void parseUsernameLength() {
         Map<String, Object> validations = valueForKey("validation", Map.class);
-        if (validations == null || !validations.containsKey("username")) {
+        if (validations == null) {
             minUsernameLength = MIN_USERNAME_LENGTH;
             maxUsernameLength = MAX_USERNAME_LENGTH;
+            return;
+        }
+        if (!validations.containsKey("username")){
+            minUsernameLength = UNUSED_USERNAME_LENGTH;
+            maxUsernameLength = UNUSED_USERNAME_LENGTH;
             return;
         }
         final Map<String, Object> usernameValidation = (Map<String, Object>) validations.get("username");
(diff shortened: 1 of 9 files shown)


lib/src/main/java/com/auth0/android/lock/internal/configuration/Connection.java
```diff
@@ -158,14 +158,19 @@ static Connection newConnectionFor(@NonNull String strategy, Map<String, Object>
 
     private void parseUsernameLength() {
         Map<String, Object> validations = valueForKey("validation", Map.class);
-        if (validations == null || !validations.containsKey("username")) {
+        if (validations == null) {
             minUsernameLength = MIN_USERNAME_LENGTH;
             maxUsernameLength = MAX_USERNAME_LENGTH;
             return;
         }
+        if (!validations.containsKey("username")){
```
hzalaz: Why not `1` and `Int.Max`?

lib/src/main/java/com/auth0/android/lock/views/ValidatedInputView.java
```diff
@@ -225,7 +225,7 @@ private void setupInputValidation() {
                 input.setInputType(InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS);
                 inputIcon = R.drawable.com_auth0_lock_ic_username;
                 hint = getResources().getString(R.string.com_auth0_lock_hint_username);
-                error = getResources().getString(R.string.com_auth0_lock_input_error_empty);
+                error = getResources().getString(R.string.com_auth0_lock_input_error_username_empty);
```
hzalaz: Is this a breaking change since we removed an error string? Also if we are doc them we should update that doc

hzalaz:
@lbalmaceda pressed the wrong button, please address the comment 😄

lib/src/main/java/com/auth0/android/lock/internal/configuration/DatabaseConnection.java
```diff
@@ -47,4 +47,11 @@
      * @return the maximum username length.
      */
     int getMaxUsernameLength();
+
+    /**
+     * Whether this connection is a Custom/Imported Database.
+     *
+     * @return true if this connection is a Custom/Imported Database.
+     */
+    boolean isCustomDatabase();
```
hzalaz: Are we testing this?

PR #373, opened 2016-11-29
Avoid sending authentication parameters on password-reset

lbalmaceda:
*Depends on https://github.com/auth0/Lock.Android/pull/374*


lib/src/main/java/com/auth0/android/lock/WebProvider.java
```diff
@@ -0,0 +1,57 @@
...
+import android.support.annotation.NonNull;
+
+import com.auth0.android.Auth0;
+import com.auth0.android.provider.WebAuthProvider;
+
+/**
+ * The WebProvider class is a wrapper for calls to the WebAuthProvider static methods.
+ * This is only for testing purposes.
+ */
+class WebProvider {
```
hzalaz: I don't see why this is needed?

hzalaz:
@lbalmaceda rebase and merge

PR #382, opened 2016-12-26
Support custom schemes for Web Authentication

lbalmaceda:
For *Android App Links* a scheme equal to `https` is required, but if you're not targeting API 23 or upper or you don't want to support Android App Links then you might want to use a custom scheme so that the user isn't prompted to chose an App to resolve the intent whenever a Web Authentication is successful. 

```java
Lock lock = lockBuilder.withScheme("auth0")
                      .build(this);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,8 @@ startActivity(lock.newIntent(this));
 ```
 
 #### Android App Links - Custom Scheme
-Currently the default scheme used for Web Authentication is `https`. This works best for Android API 23 or newer if you're using [Android App Links](https://developer.android.com/training/app-links/index.html), but in previous Android versions this may show the intent chooser dialog prompting the user to chose either your application or the browser to resolve the intent. You can change this behavior by using a custom unique scheme, so that the OS opens directly the link with your app.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -201,6 +201,14 @@ Then just start `PasswordlessLockActivity` from inside your `Activity`
 startActivity(lock.newIntent(this));
 ```
 
+#### Android App Links - Custom Scheme
+Currently the default scheme used for Web Authentication is `https`. This works best for Android API 23 or newer if you're using [Android App Links](https://developer.android.com/training/app-links/index.html), but in previous Android versions this may show the intent chooser dialog prompting the user to chose either your application or the browser to resolve the intent. You can change this behavior by using a custom unique scheme, so that the OS opens directly the link with your app.
+1. Update the Intent Filter definition in the Android Manifest and change the custom scheme.
```
hzalaz: Add a new line before to proper render the list item

README.md
```diff
@@ -201,6 +201,14 @@ Then just start `PasswordlessLockActivity` from inside your `Activity`
 startActivity(lock.newIntent(this));
 ```
 
+#### Android App Links - Custom Scheme
+Currently the default scheme used for Web Authentication is `https`. This works best for Android API 23 or newer if you're using [Android App Links](https://developer.android.com/training/app-links/index.html), but in previous Android versions this may show the intent chooser dialog prompting the user to chose either your application or the browser to resolve the intent. You can change this behavior by using a custom unique scheme, so that the OS opens directly the link with your app.
```
hzalaz: `Currently the default scheme used for Web Authentication` -> `Currently the default scheme used for the redirect url used to send the Web Auth results`

hzalaz: Also update the javadocs to state the same

PR #410, opened 2017-04-07
Send login_hint on enterprise web auth when username/email is available

lbalmaceda:
When an enterprise connection supports logging in using Web OAuth, the query parameter `login_hint` will be sent in the authorize url containing the username/email.

### DIFF
--- a/lib/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
+++ b/lib/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
@@ -73,6 +73,7 @@ import com.squareup.okhttp.OkHttpClient;
 import com.squareup.otto.Bus;
 import com.squareup.otto.Subscribe;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 
@@ -422,8 +423,8 @@ public class PasswordlessLockActivity extends AppCompatActivity implements Activ
         lastPasswordlessCountry = null;
         Log.v(TAG, "Looking for a provider to use with the connection " + event.getConnection());
         currentProvider = AuthResolver.providerFor(event.getStrategy(), event.getConnection());
-        HashMap<String, Object> authParameters = new HashMap<>(options.getAuthenticationParameters());
         if (currentProvider != null) {
+            HashMap<String, Object> authParameters = new HashMap<>(options.getAuthenticationParameters());
             final String connectionScope = options.getConnectionsScope().get(event.getConnection());
             if (connectionScope != null) {
                 authParameters.put(Constants.CONNECTION_SCOPE_KEY, connectionScope);
@@ -442,7 +443,7 @@ public class PasswordlessLockActivity extends AppCompatActivity implements Activ
         }
 
         Log.d(TAG, "Couldn't find an specific provider, using the default: " + WebAuthProvider.class.getSimpleName());
-        webProvider.start(this, event.getConnection(), authParameters, authProviderCallback, WEB_AUTH_REQUEST_CODE);
+        webProvider.start(this, event.getConnection(), null, authProviderCallback, WEB_AUTH_REQUEST_CODE);
     }
 
     //Callbacks
--- a/lib/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
+++ b/lib/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
@@ -422,8 +422,8 @@ public class PasswordlessLockActivity extends AppCompatActivity implements Activ
         lastPasswordlessCountry = null;
         Log.v(TAG, "Looking for a provider to use with the connection " + event.getConnection());
         currentProvider = AuthResolver.providerFor(event.getStrategy(), event.getConnection());
+        HashMap<String, Object> authParameters = new HashMap<>(options.getAuthenticationParameters());
         if (currentProvider != null) {
-            HashMap<String, Object> authParameters = new HashMap<>(options.getAuthenticationParameters());
             final String connectionScope = options.getConnectionsScope().get(event.getConnection());
             if (connectionScope != null) {
                 authParameters.put(Constants.CONNECTION_SCOPE_KEY, connectionScope);
@@ -442,7 +442,7 @@ public class PasswordlessLockActivity extends AppCompatActivity implements Activ
         }
 
         Log.d(TAG, "Couldn't find an specific provider, using the default: " + WebAuthProvider.class.getSimpleName());
-        webProvider.start(this, event.getConnection(), authProviderCallback, WEB_AUTH_REQUEST_CODE);
+        webProvider.start(this, event.getConnection(), authParameters, authProviderCallback, WEB_AUTH_REQUEST_CODE);
     }
 
     //Callbacks
--- a/lib/src/main/java/com/auth0/android/lock/WebProvider.java
+++ b/lib/src/main/java/com/auth0/android/lock/WebProvider.java
@@ -32,19 +32,24 @@ class WebProvider {
     /**
      * Configures a new instance of the WebAuthProvider.Builder and starts it.
      *
-     * @param activity       a valid Activity context
-     * @param connection     to use in the authentication
-     * @param authParameters extra authentication parameters to use. If null, it will use the ones defined by Options#getAuthenticationParameters
-     * @param callback       to deliver the authentication result to
-     * @param requestCode    to use in the startActivityForResult request
+     * @param activity            a valid Activity context
+     * @param connection          to use in the authentication
+     * @param extraAuthParameters extra authentication parameters to use along with the provided in the Options instance
+     * @param callback            to deliver the authentication result to
+     * @param requestCode         to use in the startActivityForResult request
      */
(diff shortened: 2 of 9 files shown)


lib/src/main/java/com/auth0/android/lock/views/LogInFormView.java
```diff
@@ -179,7 +179,14 @@ private void resetDomain() {
     }
 
     private String getUsername() {
-        return currentConnection == null && fallbackToDatabase ? emailInput.getText() : usernameInput.getText();
+        boolean usingDatabase = currentConnection == null && fallbackToDatabase;
+        boolean usingOAuthEnterprise = currentConnection != null && !currentConnection.isActiveFlowEnabled();
+        if (usingDatabase || usingOAuthEnterprise) {
+            return emailInput.getText();
+        } else {
```
lbalmaceda: This can be 1 line.

lib/src/main/java/com/auth0/android/lock/WebProvider.java
```diff
@@ -28,15 +32,19 @@
...
+     * @param activity       a valid Activity context
+     * @param connection     to use in the authentication
+     * @param authParameters extra authentication parameters to use. If null, it will use the ones defined by Options#getAuthenticationParameters
+     * @param callback       to deliver the authentication result to
+     * @param requestCode    to use in the startActivityForResult request
      */
-    public void start(Activity activity, String connection, AuthCallback callback, int requestCode) {
+    public void start(@NonNull Activity activity, @NonNull String connection, @Nullable Map<String, Object> authParameters, @NonNull AuthCallback callback, int requestCode) {
+        if (authParameters == null) {
+            authParameters = new HashMap<>(options.getAuthenticationParameters());
```
nikolaseu: I don't like this modifying/reusing a param you receive.
Also this has to be a deep copy?

lib/src/main/java/com/auth0/android/lock/WebProvider.java
```diff
@@ -28,15 +32,19 @@
...
-     * @param callback    to deliver the authentication result to
-     * @param requestCode to use in the startActivityForResult request
+     * @param activity       a valid Activity context
+     * @param connection     to use in the authentication
+     * @param authParameters extra authentication parameters to use. If null, it will use the ones defined by Options#getAuthenticationParameters
+     * @param callback       to deliver the authentication result to
+     * @param requestCode    to use in the startActivityForResult request
      */
-    public void start(Activity activity, String connection, AuthCallback callback, int requestCode) {
+    public void start(@NonNull Activity activity, @NonNull String connection, @Nullable Map<String, Object> authParameters, @NonNull AuthCallback callback, int requestCode) {
```
nikolaseu: maybe this can be "extraAuthParameters" that just extends/overrides the defaults in options?
why there are options in both places?

lib/src/main/java/com/auth0/android/lock/views/LogInFormView.java
nikolaseu: are you just making code more understandable or is this required for https://github.com/auth0/Lock.Android/pull/410/files#diff-46c7332ff70ade6b4cc8caac4e443222R234 ?

lib/src/test/java/com/auth0/android/lock/WebProviderTest.java
```diff
@@ -52,7 +58,42 @@ public void shouldStart() throws Exception {
...
+        AuthCallback callback = mock(AuthCallback.class);
+        WebProvider webProvider = new WebProvider(options);
+        Activity activity = spy(Robolectric.buildActivity(Activity.class)
+                .create()
+                .start()
+                .resume()
+                .get());
+
+        Map<String, Object> parameters = new HashMap<>();
+        parameters.put("custom-param-1", "value-1");
```
nikolaseu: are these used/checked somewhere?

lib/src/test/java/com/auth0/android/lock/WebProviderTest.java
```diff
@@ -52,7 +58,42 @@ public void shouldStart() throws Exception {
...
+        account.setOIDCConformant(true);
+        Options options = new Options();
+        options.setAccount(account);
+
+        options.setUseBrowser(true);
+        options.withAudience("https://me.auth0.com/myapi");
+
+        AuthCallback callback = mock(AuthCallback.class);
+        WebProvider webProvider = new WebProvider(options);
+        Activity activity = spy(Robolectric.buildActivity(Activity.class)
```
nikolaseu: this code is repeated a lot, maybe move to `@Before` "setUp" method?

lib/src/test/java/com/auth0/android/lock/WebProviderTest.java
```diff
@@ -40,6 +43,9 @@
     @Rule
     public ExpectedException exception = ExpectedException.none();
 
+    @Mock
+    public Map<String, Object> map;
```
nikolaseu: used only once in https://github.com/auth0/Lock.Android/pull/410/files#diff-206bfaf63e03ad1b5fec7bc2d3b5005bR61
since you're not checking anything.. can't you use null like in the other places?

lib/src/test/java/com/auth0/android/lock/LockActivityTest.java
```diff
@@ -507,4 +582,18 @@ public void shouldResumeOAuthAuthenticationWithCustomProviderOnNewIntent() throw
...
+        OAuthConnection connection = mock(OAuthConnection.class);
+        when(connection.getName()).thenReturn("my-connection");
+        OAuthLoginEvent event = new OAuthLoginEvent(connection, "user@domain.com", null);
+        activity.onOAuthAuthenticationRequest(event);
+
+        Intent intent = mock(Intent.class);
+        activity.onNewIntent(intent);
+
+        verify(lockView).showProgress(false);
+        verify(webProvider).resume(eq(intent));
```
nikolaseu: I think there's no need for `eq` here

lib/src/main/java/com/auth0/android/lock/PasswordlessLockActivity.java
```diff
@@ -442,7 +442,7 @@ public void onOAuthAuthenticationRequest(OAuthLoginEvent event) {
         }
 
         Log.d(TAG, "Couldn't find an specific provider, using the default: " + WebAuthProvider.class.getSimpleName());
-        webProvider.start(this, event.getConnection(), authProviderCallback, WEB_AUTH_REQUEST_CODE);
+        webProvider.start(this, event.getConnection(), authParameters, authProviderCallback, WEB_AUTH_REQUEST_CODE);
```
nikolaseu: since `authParameters` is not modified for usage here, just use `options.getAuthenticationParameters()` (or null) and avoid to always make the deep copy above

lib/src/main/java/com/auth0/android/lock/views/LogInFormView.java
lbalmaceda: The previous check was not considering the `activeFlow` value of the connection, so now there are 3 cases:
1. It's a db connection -> use /ro
2. It's a enterprise connection with active flow -> use /ro
3. it's a enterprise connection without active flow -> use /authorize

What I require in the last case is to provide the username/email typed before clicking submit and send as `login_hint` in the /authorize call. (This is the PR).

In this method is where I take that case and chose the right `EditText`, as in the case of (1) and (2) there are 2 steps/screens: the first one shows email and password inputs and if the typed domain matches, it will hide the password and replace the email with a username input.

lib/src/main/java/com/auth0/android/lock/WebProvider.java
lbalmaceda: The `WebProvider` class it's just an internal (pkg priv) wrapper for calls to the auth0's `WebAuthProvider`. It was created with the purpose of passing just the `Options` in the constructor and letting it handle all the params settings inside, as they weren't going to change at least for Lock's lifecycle.. 
With this PR, I need to set a value that it's different everytime the "oauth login" it's called -> the username/email to set as `login_hint`. There were 2 ways to do it:
- Add a setter for this parameter (or a generic setter for any parameter or parameters) and call it before each call to `start`. This would require to keep an extra "map" reference to store this extra values.
- Change the signature (as it's internal) and add these "extra parameters", which will default to the `options.getAuthenticationParameters()` if their value is `null`.

I did the latter, but I can do the first one if that's what you prefer. Do you think of any other options?

lib/src/main/java/com/auth0/android/lock/WebProvider.java
lbalmaceda: You're right in both. Not really for the code of this method, I can use the same instance.

lib/src/test/java/com/auth0/android/lock/WebProviderTest.java
lbalmaceda: Can't move to setup as some tests change the `options + account` properties. i.e. setting OIDC conformant, or passing the audience. Maybe there's a helper method I can move this logic to..

nikolaseu: I meant the `Activity` only

PR #422, opened 2017-07-03
Add Hosted Login Page button to the demo.

lbalmaceda:
The Web Provider requests the `https://domain.auth0.com/userinfo` audience. 

Also, a few changes were required in order to use the (still not released) WebAuthProvider that uses Chrome Custom Tabs. 

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ The next step is to create an instance of `Auth0` with your client information:
 Auth0 account = new Auth0("{YOUR_AUTH0_CLIENT_ID}", "{YOUR_AUTH0_DOMAIN}");
 ```
 
-Another way to create it is using the values defined previously in the `string.xml` file, by passing just an Android Context. The name of the keys must match the ones listed above or the constructor will throw an exception.
+Another way to create it is using the values defined previously in the `string.xml` file, by passing an Android Context. The name of the keys must match the ones listed above or the constructor will throw an exception.
 
 ```java
 Auth0 account = new Auth0(context);
@@ -184,7 +184,7 @@ public class HomeActivity extends Activity {
 }
 ```
 
-And start `LockActivity` from inside your `Activity`.
+Start `LockActivity` from inside your `Activity`.
 
 ```java
 startActivity(lock.newIntent(this));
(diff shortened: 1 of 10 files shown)


lib/build.gradle
```diff
@@ -41,7 +41,7 @@ dependencies {
     compile 'com.android.support:design:25.3.1'
     compile 'com.google.code.gson:gson:2.8.0'
     compile 'com.squareup:otto:1.3.8'
-    compile 'com.auth0.android:auth0:1.8.0'
+    compile 'com.auth0.android:auth0:1.8.0-SNAPSHOT'
```
cocojoe: What's difference?

lbalmaceda: That's pointing to the local dependency until we release 1.9.0, then CI passes 👍

README.md
```diff
@@ -238,7 +315,7 @@ public class HomeActivity extends Activity {
 }
 ```
 
-Then just start `PasswordlessLockActivity` from inside your `Activity`
+And start `PasswordlessLockActivity` from inside your `Activity`.
```
cocojoe: Might be better as "Next, start" or simply "Start"

README.md
```diff
@@ -34,25 +34,43 @@ If you haven't done yet, go to [Auth0](https://auth0.com) and create an Account,
...
+``` 
+
+The next step is to create an instance of `Auth0` with your client information:
+
+```java
+Auth0 account = new Auth0("{YOUR_AUTH0_CLIENT_ID}", "{YOUR_AUTH0_DOMAIN}");
 ```
 
-And then create a new Auth0 instance by passing an Android Context:
+Another way to create it is using the values defined previously in the `string.xml` file, by passing just an Android Context. The name of the keys must match the ones listed above or the constructor will throw an exception.
```
cocojoe: "by passing ~~just~~ an Android"

README.md
```diff
@@ -150,7 +184,7 @@ public class HomeActivity extends Activity {
 }
 ```
 
-Then just start `LockActivity` from inside your `Activity`.
+And start `LockActivity` from inside your `Activity`.
```
cocojoe: It looks strange in English to start a sentence with "And" So I would juts use "Start"

PR #434, opened 2017-08-09
Update custom scheme usage instructions on the README.md

lbalmaceda:


README.md
```diff
@@ -138,7 +138,7 @@ In case you are using an older version of Lock for **Social** Authentication, th
 Make sure the Activity's `launchMode` is declared as `singleTask` or the result won't come back in the authentication.
 
 
-Then in any of your Activities you need to initialize **Lock**:
+Then in any of your Activities you need to initialize **Lock**. In case that your `auth0Scheme` value is different from `https` you should tell the library which scheme to use by calling `withScheme()` as pointed below.
```
cocojoe: `If your `auth0Scheme` value is not `https` then you need to use `withScheme()` as shown below.

README.md
```diff
@@ -269,7 +270,7 @@ When the Passwordless connection is SMS you must also add the `CountryCodeActivi
 ```
 
 
-Then in any of your Activities you need to initialize **PasswordlessLock**
+Then in any of your Activities you need to initialize **PasswordlessLock**. In case that your `auth0Scheme` value is different from `https` you should tell the library which scheme to use by calling `withScheme()` as pointed below.
```
cocojoe: `If your `auth0Scheme` value is not `https` then you need to use `withScheme()` as shown below.

README.md
```diff
@@ -322,11 +324,11 @@ startActivity(lock.newIntent(this));
 ```
 
 #### Android App Links - Custom Scheme
-Currently the default scheme used for the redirect url used to send the Web Auth results is `https`. This works best for Android Marshmallow (API 23) or newer if you're using [Android App Links](https://developer.android.com/training/app-links/index.html), but in previous Android versions this may show the intent chooser dialog prompting the user to chose either your application or the browser to resolve the intent. You can change this behavior by using a custom unique scheme, so that the OS opens the link directly with your app.
+The current default scheme used by the library to generate the Redirect URL for Web Authentication is `https`. This works best for Android Marshmallow (API 23) or newer if you're using [Android App Links](https://developer.android.com/training/app-links/index.html), but in previous Android versions this may show the intent chooser dialog prompting the user to choose either your application or the browser to resolve the intent. You can change this behavior by using a custom unique scheme, so that the OS opens the link directly with your app.
```
cocojoe: , but in previous -> .However, in previous

PR #435, opened 2017-08-14
Add Change Password screen

lbalmaceda:
This PR adds support for the **Self Change Password** endpoint.

We need to tweak the copy/text of the description.
image

image

Both "new passwords" must match. Validation runs while typing the second field or when the form is submitted:
image

Focus on the green message and ignore the screen below. After a successful request, lock will auto login the user using the `identity` and the `newPassword` while showing the Login screen.
image

Password policy. We need to tweak the margin between the description and the password policy part of the widget. Also, because of the narrow description space, the form expands too much and it's made scrollable.
image

If allowShowPassword is `true` the password confirm field is hidden.
image

image


cocojoe:
@lbalmaceda do we still actually want this? If not should remove review request and close.

Issue #425, opened 2017-07-06
Google or Facebook buttons disappear in Right-to-Left language layouts


lidiakrus:
We are using Auth0 for Android that allows user to log in with Google or Facebook. 

The translations for the login dialog such as e.g. Spanish, Russian, Korean all worked fine.

Arabic UI test (with a couple of translated strings) has shown, that Facebook and Google login buttons disappeared. (The translations themselves worked fine). 
Then I just changed the device language to Urdu without providing any translations. Same issue there - Google and Facebook buttons disappeared. (This urdu version is demonstrated in the screenshot in the middle)

So my guess is, this is issue with RTL languages layout. 

Or maybe I am doing something wrong?

Hope this is a correct place to ask about this issue.

![langs](https://user-images.githubusercontent.com/4237779/27915837-ad0f5a2c-6267-11e7-93cd-791a81ee0fc6.png)

- [2.5.0] Lock version
- [Samsung Galaxy S7 - Android 7.0 or any other of our test devices] Android version & Phone Model when appropriate
- [Build tools - 26.0.0, sdk 26] Versions of Gradle, Android Plugin, Build tools & Android SDK

(I am currently building with sdk 26, but I have tested an older apk that was built with sdk 25, and the issue is the same.)

lbalmaceda:
@lidiakrus thanks for the detailed report. We're aware that Lock's layout is not RTL optimized and will fix it in a future release. Will leave this open as a reminder.
Cheers. 🎉

lidiakrus:
good to know! thanks!

PR #445, opened 2017-10-12
Add RTL support

lbalmaceda:
This PR adds RTL languages support by flipping views horizontally. It also rewrites the `ValidatedInputView` layout and makes thinner the outline. (2dp->1dp)

Sample output (forced RTL):
image

Fixes https://github.com/auth0/Lock.Android/issues/425


cocojoe:
@lbalmaceda why do you need the extra assets? you can't just flip the ones you have when in RTL mode?

lbalmaceda:
@cocojoe some layouts are automatically flipped but some others don't, like our custom views. Those have to be flipped manually (code/xml). Regarding the assets, I've spoken with design and they told me to flip the ones that "point" to a direction, which in our case are the chevrons.

lbalmaceda:
@cocojoe walkthrough http://recordit.co/ciiLyNqDMQ :)

PR #474, opened 2018-08-24
Allow to override the password minimum length

lbalmaceda:
Each database connection can specify a password policy to apply to the password field validation (e.g. on sign up). Additionally, a minimum password length can be specified to override the policy's minimum. This information arrives from the CDN file (see `ApplicationFetcher` class).

**Note to the reviewer:** Classes inside the `internal` package can contain breaking changes. I've created a new class `PasswordComplexity` to store this new value and any other future values that are introduced for password validation. I've kept the classes inside the `view` package without breaking changes just in case 🙆‍♂️ 

#### Screenshots
Using Policy "Low" (min 3 chars) and changing the minimum to 4 and 8

image
image

image

image

### DIFF
--- a/lib/src/main/java/com/auth0/android/lock/internal/configuration/Configuration.java
+++ b/lib/src/main/java/com/auth0/android/lock/internal/configuration/Configuration.java
@@ -273,9 +273,9 @@ public class Configuration {
         return passwordlessMode;
     }
 
-    @PasswordStrength
-    public int getPasswordPolicy() {
-        return defaultDatabaseConnection == null ? PasswordStrength.NONE : defaultDatabaseConnection.getPasswordPolicy();
+    @NonNull
+    public PasswordComplexity getPasswordComplexity() {
+        return defaultDatabaseConnection == null ? new PasswordComplexity(PasswordStrength.NONE, null) : defaultDatabaseConnection.getPasswordComplexity();
     }
 
     public boolean loginAfterSignUp() {
--- a/lib/src/main/java/com/auth0/android/lock/internal/configuration/Connection.java
+++ b/lib/src/main/java/com/auth0/android/lock/internal/configuration/Connection.java
@@ -19,6 +19,7 @@ public class Connection implements BaseConnection, DatabaseConnection, OAuthConn
     private int maxUsernameLength;
     private boolean isCustomDatabase;
     private boolean allowActiveFlow = true;
+    private PasswordComplexity passwordComplexity;
 
     private Connection(@NonNull String strategy, Map<String, Object> values) {
         checkArgument(values != null && values.size() > 0, "Must have at least one value");
@@ -28,6 +29,31 @@ public class Connection implements BaseConnection, DatabaseConnection, OAuthConn
         this.name = name;
         this.values = values;
         parseUsernameLength();
+        parsePasswordComplexity();
+    }
+
+    private void parsePasswordComplexity() {
+        int policy = PasswordStrength.NONE;
+        String value = valueForKey("passwordPolicy", String.class);
+        if ("excellent".equals(value)) {
+            policy = PasswordStrength.EXCELLENT;
+        }
+        if ("good".equals(value)) {
+            policy = PasswordStrength.GOOD;
+        }
+        if ("fair".equals(value)) {
+            policy = PasswordStrength.FAIR;
+        }
+        if ("low".equals(value)) {
+            policy = PasswordStrength.LOW;
+        }
+
+        final Map<String, Object> complexityOptions = valueForKey("password_complexity_options", Map.class);
+        Integer minLength = null;
+        if (complexityOptions != null && complexityOptions.containsKey("min_length")) {
+            minLength = ((Number) complexityOptions.remove("min_length")).intValue();
+        }
+        passwordComplexity = new PasswordComplexity(policy, minLength);
     }
 
     @Override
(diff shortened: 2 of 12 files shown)


lib/src/main/java/com/auth0/android/lock/internal/configuration/Configuration.java
```diff
@@ -273,9 +273,9 @@ public int getPasswordlessMode() {
         return passwordlessMode;
     }
 
-    @PasswordStrength
-    public int getPasswordPolicy() {
```
joshcanhelp: Change to a public API .... BC?

lib/src/main/java/com/auth0/android/lock/internal/configuration/Connection.java
```diff
@@ -28,6 +29,31 @@ private Connection(@NonNull String strategy, Map<String, Object> values) {
         this.name = name;
         this.values = values;
         parseUsernameLength();
+        parsePasswordComplexity();
+    }
+
+    private void parsePasswordComplexity() {
+        int policy = PasswordStrength.NONE;
+        String value = valueForKey("passwordPolicy", String.class);
+        if ("excellent".equals(value)) {
```
joshcanhelp: Is a `switch` a better fit here? Seems like checking the same thing over and over is duplicative.

lib/src/main/java/com/auth0/android/lock/internal/configuration/DatabaseConnection.java
```diff
@@ -1,17 +1,19 @@
...
-     * Getter for the Password Policy associated to this connection.
+     * Getter for the Password Validation settings
      *
-     * @return The Password Policy level for this connection.
+     * @return The Password Validation for this connection.
      */
-    @PasswordStrength
-    int getPasswordPolicy();
+    @NonNull
+    PasswordComplexity getPasswordComplexity();
```
joshcanhelp: Seems like another change to a public API.

lib/src/main/java/com/auth0/android/lock/views/PasswordStrengthView.java
```diff
@@ -181,9 +186,20 @@ private int getMinimumLength() {
      * Sets the current level of Strength that this widget is going to validate.
      *
      * @param strength the required strength level.
+     * @deprecated use the {@link #setPasswordComplexity(PasswordComplexity)} method
```
joshcanhelp: Separate PR for deprecations?

lib/src/main/java/com/auth0/android/lock/internal/configuration/PasswordComplexity.java
```diff
@@ -0,0 +1,34 @@
...
+     */
+    @PasswordStrength
+    public int getPasswordPolicy() {
+        return policy;
+    }
+
+    /**
+     * Getter for the minimum length the password requires
+     *
+     * @return the minimum length the password requires
```
joshcanhelp: Should doc comments like this start with a capital letter or not? You're mixing case in this class.

lib/src/main/java/com/auth0/android/lock/views/ValidatedPasswordInputView.java
```diff
@@ -53,8 +54,19 @@ protected void updateBorder() {
      * Sets the Password Strength Policy level for this view.
      *
      * @param strength the new Policy level.
+     * @deprecated use {@link #setPasswordComplexity(PasswordComplexity)} method
```
joshcanhelp: Separate PR for deprecations?

lib/src/test/java/com/auth0/android/lock/views/PasswordStrengthViewTest.java
```diff
@@ -189,4 +190,59 @@ public void shouldHandlePasswordStrengthEXCELLENT() throws Exception {
         assertFalse(view.isValid(PASSWORD_TOO_LONG));
         assertFalse(view.isValid(null));
     }
+
+    @Test
+    public void shouldOverrideMinLength() throws Exception {
```
joshcanhelp: Nice work in here 👍 super-clear

lib/src/main/java/com/auth0/android/lock/internal/configuration/Configuration.java
lbalmaceda: classes under the `internal` package can change. There's a header on (some of) the files explaining that.

lib/src/main/java/com/auth0/android/lock/views/PasswordStrengthView.java
lbalmaceda: Seems an overkill since people shouldn't be using our view classes directly.

lib/src/main/java/com/auth0/android/lock/internal/configuration/PasswordComplexity.java
lbalmaceda: when talking about params or return values I rather start with a lower case.

lib/src/main/java/com/auth0/android/lock/internal/configuration/PasswordComplexity.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public PasswordComplexity(@PasswordStrength int policy, @Nullable Integer minLengthOverride) {
+        this.policy = policy;
+        this.minLengthOverride = minLengthOverride;
+    }
+
+    /**
+     * Getter for the Password Policy associated to this connection.
+     *
+     * @return The Password Policy level for this connection.
```
joshcanhelp: @lbalmaceda - Based on your other comment, this should be lowercase

PR #478, opened 2018-09-19
Don't show the signup terms banner if there is no terms string

luisrudge:
We're making this change to make all Lock versions behave in the same way.

The idea here is that the Terms and Conditions banner will be always visible, unless the translated string `com_auth0_lock_sign_up_terms` is empty.

To hide it, just set `com_auth0_lock_sign_up_terms` to an empty string in the internationalization file.

Default:
image

With an empty `com_auth0_lock_sign_up_terms` string:
image


cocojoe:
Superseded by #487

PR #479, opened 2018-09-26
Upgrade to API and Support Library 28

ligol:
I've upgraded the project to work on api 28 with the support library 28 to get it ready for AndroidX.
With the actual version Jetifier have a issue with the setOnTabSelectedListener who is now deprecated.

To make this change I needed to commet the oss section in gradle due to the plugin forcing a target api 25

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,13 +1,13 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 25
-    buildToolsVersion '25.0.2'
+    compileSdkVersion 28
+    buildToolsVersion '28.0.2'
 
     defaultConfig {
         applicationId "com.auth0.android.lock.app"
         minSdkVersion 15
-        targetSdkVersion 25
+        targetSdkVersion 28
         versionCode 1
         versionName "1.0"
         manifestPlaceholders = [auth0Domain: "@string/com_auth0_domain", auth0Scheme: "demo"]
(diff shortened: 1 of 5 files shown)


lbalmaceda:
@ligol I appreciate the effort but:
- This breaks our release plugin, which needs some love as well.
- This was [_just_ released](https://android-developers.googleblog.com/2018/09/android-studio-32.html) and even if it's already "stable" I'd rather wait a while for the bug hunt.

I'll close this for the time being and I hope that in the future I can spend some time on updating the release plugin so we can make an update on this SDK like the one you did on this PR.

ligol:
Could it be at least possible to update to a version where the method addOnTabSelectedListener method was added(26.0.1), I will normally be enough to get Jetifier to work with the library and don't crash when you open the lock screen for a ClassNotFoundException

lbalmaceda:
@ligol I don't see how the dependencies on this library can affect the ones on your project. You should require the dependencies you need in the version you need, independently from what it's required by this SDK. And gradle should be smart enough to manage the dependencies on its own. Can you add more details?

lbalmaceda:
That method did not get deprecated but **removed**. If it were deprecated one would still be able to call it just fine. They seem to have removed it in 26 so I'm confident that maybe just by updating this SDK to use support libraries version 26 your issue will be fixed. I'll review this 👍

ligol:
Where did you see it was removed, they say deprecated on the doc and it's still present in the version 28

https://developer.android.com/reference/android/support/design/widget/TabLayout#setontabselectedlistener

lbalmaceda:
My point is, if the method wasn't removed you shouldn't be getting a `MethodNotFoundException` when you use the SDK as it is today. Unless of course the `MethodNotFoundException` is caused by something else.

lbalmaceda:
@ligol I've just released `2.11.0` targeting 28. See if that works fine for you. I did not encounter the `setOnTabSelectedListener` method error.

PR #480, opened 2018-10-01
allow the user to enable logs in the builder

SansBailmare:


lbalmaceda:
@SansBailmare this is already available. You need to call [this method](https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/Auth0.java#L204) on the Auth0 instance you pass to the builder constructor.

PR #481, opened 2018-10-01
Replaced a deprecated method call with recommended

overpas:
A call of the method setOnTabSelectedListener in TabLayout doesn't work well with latest versions of Android support library as this method has been deprecated.


lbalmaceda:
@overpas Should be solved in latest release. Otherwise, please leave some feedback 👍

PR #482, opened 2018-10-01
Replace compile with implementation on README

rosariopfernandes:
Most recent versions of gradle now use `implementation` instead of `compile`


lbalmaceda:
Already done on #484

PR #487, opened 2018-10-04
Added showTerms feature flag

cocojoe:
**Description**

Added a `showTerms` feature flag to disable the display of the terms of service dialog. Displayed in the Sign Up Screen for Database connections.
Note: If `mustAcceptTerms` is enabled, terms will always be shown.

screenshot_1538661279

### DIFF
--- a/lib/src/main/java/com/auth0/android/lock/Lock.java
+++ b/lib/src/main/java/com/auth0/android/lock/Lock.java
@@ -568,8 +568,8 @@ public class Lock {
         }
 
         /**
-         * Displays the Privacy Policy and Terms of Service dialog.
-         * Note: The dialog will always be shown if the mustAcceptTerms flag has been enabled.
+         * Displays the Privacy Policy and Terms of Service footer on the Sign Up screen.
+         * Note: The footer will always be shown if the mustAcceptTerms flag has been enabled.
          * The default value is true.
          *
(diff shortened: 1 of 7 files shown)


lib/src/main/java/com/auth0/android/lock/Lock.java
```diff
@@ -567,6 +567,19 @@ public Builder setMustAcceptTerms(boolean mustAcceptTerms) {
             return this;
         }
 
+        /**
+         * Displays the Privacy Policy and Terms of Service dialog.
```
lbalmaceda: "Displays the Privacy Policy and Terms of Service footer on the Sign Up screen"

cocojoe:
@lbalmaceda can you rephrase your question please, what do you mean into the custom fields sign up screen? Are you talking about how it looks with custom fields and/or how it looks if you set `initial screen to signup` ?

cocojoe:
screenshot_1538666109
screenshot_1538666129

lbalmaceda:
@cocojoe when your lock configuration has more than 3 fields the remaining fields are displayed in a separate screen, following the initial sign up screen. On that screen that I remember, the terms are not shown. With the changes on the code you added here it might be interesting checking that that use case still works with and without this new flag enabled

cocojoe:
screenshot_1538743686

PR #489, opened 2018-10-05
newer version

doilio:
Replaced
`compile` with `implementation`
`testCompile` with `testImplementation`

The compile configuration is now deprecated and should be replaced by implementation or api
From the  [Gradle documentation](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -37,9 +37,9 @@ android {
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


lbalmaceda:
Already done on #484

PR #492, opened 2018-10-08
Remove dead code

gy741:
Hello.

I think, Value stored to 'TAG' is never read 

what do you think?

Thanks.


lbalmaceda:
@gy741 sounds fair. Would you mind rebasing the branch?

gy741:
Like this?

PR #502, opened 2018-10-25
Added readability to readme.

Frenchtoast0:
Replaced a couple of "In case" with a smoother "if".  There are a few commas that can be removed according to grammar rules, however, I don't know if they are meant as a pause to make the sentence sound better.


joshcanhelp:
We appreciate the effort here but these changes are not necessary ... thank you regardless!

PR #506, opened 2018-10-29
updated minSdkVersion to 26

ritwik787:
Google has instructed that all the new Android apps on Google Play should target API Level 26 (Android Oreo) in order to be published.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -7,7 +7,7 @@ android {
 
     defaultConfig {
         applicationId "com.auth0.android.lock.app"
-        minSdkVersion 15
+        minSdkVersion 26
         targetSdkVersion 28
         versionCode 1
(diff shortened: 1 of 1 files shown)


lbalmaceda:
@ritwik787 hey! there are several things here:
- Changing the **minimum** SDK version would mean people with a version older than that one can't use your app.
- In this case, since this is a library, it doesn't matter what we choose as the only value that will matter is the one defined in your app's manifest (the app using this library as a dependency). 
- I understand that you have changed the _demo_ app but still, we don't publish it to the playstore and if we would, the change should be on the **target** SDK version and not the minimum SDK version, as explained in the official blogpost [here](https://developer.android.com/distribute/best-practices/develop/target-sdk).  

I'll close this as it's not required anyway. Cheers

PR #522, opened 2019-02-22
Parse 'password_leaked' error message

lbalmaceda:
Add the ability to parse `password_leaked` errors returned by the backend. This is currently handled in a generic way, displaying a default error message when this case happens. 

If this PR gets merged, the error would look like this instead:
image

Copy subject to change and open to suggestions :) 

> The string resource can be overwritten by defining your own string resource using the very same key.


lbalmaceda:
Changed the copy as suggested internally 
image

PR #529, opened 2019-04-26
Remove "Small" social button style

lbalmaceda:
### Changes

The small social button style is no longer offered since this. In this PR I'm removing the logic to change it, but keeping the methods in order to avoid any code breaking changes. There will be a visual change as from now on all social buttons will appear as large/big with their connection name rather than small without text.

### Testing

Tests have been updated to reflect the change. New default or value is always BIG.

### Checklist

### DIFF
--- a/lib/src/main/java/com/auth0/android/lock/internal/configuration/Configuration.java
+++ b/lib/src/main/java/com/auth0/android/lock/internal/configuration/Configuration.java
@@ -260,6 +260,7 @@ public class Configuration {
         return initialScreen;
     }
 
+    @Deprecated
     @AuthButtonSize
     public int getSocialButtonStyle() {
         return socialButtonStyle;
--- a/lib/src/main/java/com/auth0/android/lock/internal/configuration/Options.java
+++ b/lib/src/main/java/com/auth0/android/lock/internal/configuration/Options.java
@@ -64,7 +64,6 @@ public class Options implements Parcelable {
     private boolean useBrowser;
     private boolean usePKCE;
     private boolean closable;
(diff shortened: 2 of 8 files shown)


lib/src/main/java/com/auth0/android/lock/internal/configuration/Configuration.java
```diff
@@ -260,6 +260,7 @@ public int getInitialScreen() {
         return initialScreen;
     }
 
+    @Deprecated
```
lbalmaceda: internal class, no need to document

lib/src/main/java/com/auth0/android/lock/internal/configuration/Options.java
```diff
@@ -335,13 +332,14 @@ public void setAllowLogIn(boolean allowLogIn) {
...
+    @Deprecated
     public void setAuthButtonSize(@AuthButtonSize int authButtonSize) {
-        this.authButtonSize = authButtonSize;
     }
 
+    @Deprecated
     @AuthButtonSize
     public int authButtonSize() {
-        return authButtonSize;
+        return AuthButtonSize.BIG;
```
lbalmaceda: hardcoded to always return BIG. 
internal class, no need to document

damieng:
Do we need to document this in the README/release notes/migration guide?

lbalmaceda:
@damieng I think we should add it somewhere, yes. If this is going to be in the changelog that will need to wait until the release. There is no migration guide in this repo. Maybe in the readme for now? What did web lock end up doing?

PR #537, opened 2019-06-06
Remove entirely the small social button style

lbalmaceda:
### Changes

Small button style was still hardcoded on some Lock configurations. This change removes entirely the definition of "small button style" and shows always the large social button.

There is no breaking change for the users.

### Testing

image

### Checklist


lib/src/main/java/com/auth0/android/lock/views/FormLayout.java
```diff
@@ -137,7 +137,7 @@ private void displayInitialScreen() {
     }
 
     private void addSocialLayout() {
-        socialLayout = new SocialView(lockWidget, true);
+        socialLayout = new SocialView(lockWidget, false);
```
lbalmaceda: Param is no longer used, but the class is public and can't change the signature. Moved to "false" to be more explicit.

lib/src/main/java/com/auth0/android/lock/views/SocialButton.java
```diff
@@ -31,16 +30,15 @@ public SocialButton(Context context, AttributeSet attrs) {
         init();
     }
 
-    public SocialButton(Context context, boolean smallSize) {
+    public SocialButton(Context context) {
```
lbalmaceda: Class is scoped as "protected" -> no breaking change

PR #543, opened 2019-07-11
Add Root Attributes on sign up or user creation

lbalmaceda:
### Changes

Allow to send user root profile attributes on sign up or user creation. 
Users can now choose when they want to store the custom fields. This still conserves the default of using the `user_metadata` key when not specified.

Docs PR: https://github.com/auth0/docs/pull/7954

### Testing

Pending manual testing. Added tests to the new static helper method that was used from 2 classes. The name can be changed, though. But people shouldn't be using our internal views in a standalone way anyway.

### Checklist

### DIFF
--- a/lib/src/main/java/com/auth0/android/lock/events/DatabaseSignUpEvent.java
+++ b/lib/src/main/java/com/auth0/android/lock/events/DatabaseSignUpEvent.java
@@ -34,16 +34,13 @@ import com.auth0.android.authentication.request.DatabaseConnectionRequest;
 import com.auth0.android.authentication.request.SignUpRequest;
 import com.auth0.android.result.DatabaseUser;
 
-import java.util.HashMap;
 import java.util.Map;
 
 public class DatabaseSignUpEvent extends DatabaseEvent {
 
-    private static final String KEY_USER_METADATA = "user_metadata";
-
     @NonNull
     private String password;
-    private Map<String, String> extraFields;
+    private Map<String, Object> extraFields;
 
     public DatabaseSignUpEvent(@NonNull String email, @NonNull String password, @Nullable String username) {
         super(email, username);
@@ -55,7 +52,7 @@ public class DatabaseSignUpEvent extends DatabaseEvent {
         return password;
     }
 
-    public void setExtraFields(@NonNull Map<String, String> customFields) {
+    public void setExtraFields(@NonNull Map<String, Object> customFields) {
         this.extraFields = customFields;
     }
 
@@ -67,9 +64,7 @@ public class DatabaseSignUpEvent extends DatabaseEvent {
             request = apiClient.signUp(getEmail(), getPassword(), connection);
         }
         if (extraFields != null) {
-            Map<String, Object> params = new HashMap<>();
-            params.put(KEY_USER_METADATA, extraFields);
-            request.addSignUpParameters(params);
+            request.addSignUpParameters(extraFields);
         }
         return request;
     }
@@ -82,7 +77,7 @@ public class DatabaseSignUpEvent extends DatabaseEvent {
             request = apiClient.createUser(getEmail(), getPassword(), connection);
         }
         if (extraFields != null) {
-            request.addParameter(KEY_USER_METADATA, extraFields);
+            request.addParameters(extraFields);
         }
         return request;
     }
--- a/lib/src/main/java/com/auth0/android/lock/events/DatabaseSignUpEvent.java
+++ b/lib/src/main/java/com/auth0/android/lock/events/DatabaseSignUpEvent.java
@@ -34,17 +34,21 @@ import com.auth0.android.authentication.request.DatabaseConnectionRequest;
 import com.auth0.android.authentication.request.SignUpRequest;
 import com.auth0.android.result.DatabaseUser;
 
+import java.util.HashMap;
 import java.util.Map;
 
 public class DatabaseSignUpEvent extends DatabaseEvent {
 
+    private static final String KEY_USER_METADATA = "user_metadata";
+
     @NonNull
     private String password;
-    private Map<String, Object> extraFields;
+    private Map<String, Object> rootAttributes;
 
     public DatabaseSignUpEvent(@NonNull String email, @NonNull String password, @Nullable String username) {
         super(email, username);
         this.password = password;
+        this.rootAttributes = new HashMap<>();
     }
 
     @NonNull
@@ -52,9 +56,19 @@ public class DatabaseSignUpEvent extends DatabaseEvent {
         return password;
     }
 
-    public void setExtraFields(@NonNull Map<String, Object> customFields) {
-        this.extraFields = customFields;
+    public void setRootAttributes(@NonNull Map<String, Object> attributes) {
+        this.rootAttributes.putAll(attributes);
     }
+
+    /**
+     * Set the fields to set as user_metadata
+     *
+     * @param customFields user_metadata fields to set
+     */
+    public void setExtraFields(@NonNull Map<String, String> customFields) {
+        this.rootAttributes.put(KEY_USER_METADATA, customFields);
+    }
+
 
     public SignUpRequest getSignUpRequest(AuthenticationAPIClient apiClient, String connection) {
         SignUpRequest request;
@@ -63,8 +77,8 @@ public class DatabaseSignUpEvent extends DatabaseEvent {
         } else {
             request = apiClient.signUp(getEmail(), getPassword(), connection);
         }
-        if (extraFields != null) {
-            request.addSignUpParameters(extraFields);
+        if (!rootAttributes.isEmpty()) {
+            request.addSignUpParameters(rootAttributes);
         }
         return request;
     }
@@ -76,8 +90,8 @@ public class DatabaseSignUpEvent extends DatabaseEvent {
         } else {
             request = apiClient.createUser(getEmail(), getPassword(), connection);
         }
-        if (extraFields != null) {
-            request.addParameters(extraFields);
+        if (!rootAttributes.isEmpty()) {
+            request.addParameters(rootAttributes);
         }
         return request;
     }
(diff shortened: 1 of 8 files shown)


lib/src/main/java/com/auth0/android/lock/events/DatabaseSignUpEvent.java
```diff
@@ -55,7 +52,7 @@ public String getPassword() {
         return password;
     }
 
-    public void setExtraFields(@NonNull Map<String, String> customFields) {
+    public void setExtraFields(@NonNull Map<String, Object> customFields) {
```
lbalmaceda: The only "breaking change" is a change in a signature, but goes from specific (String) to generic (Object) so, there shouldn't be any breaking change here. @jimmyjames please bless this :D .

lib/src/main/java/com/auth0/android/lock/utils/CustomField.java
```diff
@@ -56,22 +58,66 @@
         int TYPE_EMAIL = 3;
     }
 
+    /**
+     * Location to store the field into.
+     */
+    @IntDef({PROFILE_ROOT, USER_METADATA})
+    @Retention(RetentionPolicy.SOURCE)
+    public @interface Storage {
```
lbalmaceda: proposed "Storage" but can be named differently. Is like an interface users could use as "constant" when requiring this type of value. So users would pass `Storage.PROFILE_ROOT` instead of 0.

Should this be named something different?

lib/src/main/java/com/auth0/android/lock/utils/CustomField.java
```diff
@@ -56,22 +58,66 @@
...
+
+    /**
+     * Constructor for CustomField instance
+     *
+     * @param icon the icon resource to display next to the field.
+     * @param type the type of input this field will receive. Used to determine the applied validation.
+     * @param key  the key to store this field as.
+     * @param hint the placeholder to display when this field is empty.
+     */
     public CustomField(@DrawableRes int icon, @FieldType int type, @NonNull String key, @StringRes int hint) {
```
lbalmaceda: Should I deprecate this existing constructor that doesn't take the `Storage`? I mean.. it still works with the defaults (user metadata) so no real justification.

lib/src/main/java/com/auth0/android/lock/views/SignUpFormView.java
```diff
@@ -111,15 +110,6 @@ private void addCustomFields(List<CustomField> customFields) {
         }
     }
 
-    private Map<String, String> getCustomFieldValues() {
```
lbalmaceda: This was duped on another class. Kept on that one as a static public method. Btw, should that method `convertFieldsToMap` be renamed, now that is public?

lib/src/main/java/com/auth0/android/lock/views/CustomFieldsFormView.java
```diff
@@ -106,20 +109,30 @@ private void addCustomFields() {
         }
     }
 
-    private Map<String, String> getCustomFieldValues() {
-        Map<String, String> map = new HashMap<>();
-        for (CustomField data : fieldsData) {
-            map.put(data.getKey(), data.findValue(fieldContainer));
-        }
-        Log.d(TAG, "Custom field values are" + map.values().toString());
+    public static Map<String, Object> convertFieldsToMap(List<CustomField> fields, ViewGroup container) {
```
lbalmaceda: new method conserving the metadata logic and adding root elements as well.

lib/src/main/java/com/auth0/android/lock/events/DatabaseSignUpEvent.java
jimmyjames: It would be a breaking change, since `Map<String, String>` is actually not a subclass of `Map<String, Object>`. So given the following code:

```java
Map<String, String> m1 = new HashMap<>();
m1.put("key", "val");
dse.setExtraFields(m1);
```

Would result in a compilation failure like
```
error: incompatible types: Map<String,String> cannot be converted to Map<String,Object>
```

[This article](https://nofluffjuststuff.com/magazine/2016/09/time_to_really_learn_generics_a_java_8_perspective#object_collection) contains more information about Collections and type invariance.

jimmyjames: To make this backwards-compatible, I think you could use a bounded wildcard of `? extends Object`:

```java
public void setExtraFields(@NonNull Map<String, ? extends Object> customFields) {
```

lbalmaceda: I'll add a separate method 👍

lib/src/main/java/com/auth0/android/lock/utils/CustomField.java
```diff
@@ -56,22 +58,66 @@
...
+    public @interface Storage {
+        /**
+         * Store the field into the user profile root.
+         * The list of attributes that can be added to your root profile is here https://auth0.com/docs/api/management/v2#!/Users/patch_users_by_id.
+         */
+        int PROFILE_ROOT = 0;
+
+        /**
+         * Store the field into the user_metadata object.
+         * The list of attributes that can be added to your root profile is here https://auth0.com/docs/api/management/v2#!/Users/patch_users_by_id.
```
cocojoe: Would remove this line, not relevant for metadata.

lib/src/main/java/com/auth0/android/lock/utils/CustomField.java
```diff
@@ -56,22 +58,66 @@
...
+     * Location to store the field into.
+     */
+    @IntDef({PROFILE_ROOT, USER_METADATA})
+    @Retention(RetentionPolicy.SOURCE)
+    public @interface Storage {
+        /**
+         * Store the field into the user profile root.
+         * The list of attributes that can be added to your root profile is here https://auth0.com/docs/api/management/v2#!/Users/patch_users_by_id.
+         */
+        int PROFILE_ROOT = 0;
```
cocojoe: Shouldn't metaData be 0 as it's the original and default with root being the new one.

lib/src/main/java/com/auth0/android/lock/utils/CustomField.java
cocojoe: Your decision, personally I think it's fine as is, it's not like there is a requirement for people to move to root storage.

lib/src/main/java/com/auth0/android/lock/events/DatabaseSignUpEvent.java
```diff
@@ -43,33 +43,42 @@
...
+    }
+
+    /**
+     * Set the fields to set as user_metadata
+     *
+     * @param customFields user_metadata fields to set
+     */
     public void setExtraFields(@NonNull Map<String, String> customFields) {
-        this.extraFields = customFields;
+        this.rootAttributes.put(KEY_USER_METADATA, customFields);
```
cocojoe: You store your metadata inside the rootAttributes dictionary?

lbalmaceda: What I call here `rootAttributes` is a map of additional sign up parameters. And yes, metadata goes into them just like it was being sent before. The only difference is that before, metadata was the only attribute we would send. Now it has siblings :)

lib/src/main/java/com/auth0/android/lock/utils/CustomField.java
lbalmaceda: I agree. Probably copy-pasta

lib/src/main/java/com/auth0/android/lock/utils/CustomField.java
lbalmaceda: No need to swap these, but I was using the wrong constant value on the existing constructor so thanks for that 👍

lbalmaceda: Reference https://github.com/auth0/Lock.Android/pull/543/files#diff-2c5ac7c3a0fefe698d1137e6668dc1d1R100

lib/src/main/java/com/auth0/android/lock/utils/CustomField.java
```diff
@@ -56,22 +58,68 @@
...
+        /**
+         * Store the field into the user profile root.
+         * The list of attributes that can be added to your root profile is here https://auth0.com/docs/api/authentication#signup.
+         */
+        int PROFILE_ROOT = 0;
+
+        /**
+         * Store the field into the user_metadata object.
+         */
+        int USER_METADATA = 1;
```
cocojoe: Shouldn't USER_METADATA be default as that's existing behaviour?

lbalmaceda: It is the default https://github.com/auth0/Lock.Android/pull/543/files#diff-2c5ac7c3a0fefe698d1137e6668dc1d1R99

lib/src/main/java/com/auth0/android/lock/utils/CustomField.java
jimmyjames: `Storage` makes sense to me in this context (and the accompanying docs change reads well with `Storage`)

lib/src/main/java/com/auth0/android/lock/utils/CustomField.java
```diff
@@ -56,22 +58,68 @@
...
+    }
+
+    /**
+     * Constructor for CustomField instance
+     *
+     * @param icon    the icon resource to display next to the field.
+     * @param type    the type of input this field will receive. Used to determine the applied validation.
+     * @param key     the key to store this field as.
+     * @param hint    the placeholder to display when this field is empty.
+     * @param storage the location to store this value into. Defaults to {@link Storage#USER_METADATA}.
```
jimmyjames: How does this default to `Storage.USER_METADATA`? I see the other constructor will pass in `USER_METADATA`, but since this is also a public constructor, there really isn't a "default" here is there?

lbalmaceda: I agree. I'll clarify that the default applies to the other constructor only.

PR #549, opened 2019-09-13
Options now pass additional user meta data

morganlutz:
### Changes

This adds the option to pass-through ~~custom root attributes and~~ user metadata to the database sign up event. This enables apps to assign values to the new user that should not be displayed as user-editable `CustomField`s.

- Adds ~~`customSignUpRootAttributes` and~~ `customSignUpUserMetadata` properties on the Lock Options type.
- Passes the ~~`customSignUpRootAttributes` and~~ `customSignUpUserMetadata` values to the `DatabaseSignUpEvent` in `LockActivity#onDatabaseAuthenticationRequest`.

### References

- [Set Metadata Properties on Creation](https://auth0.com/docs/users/guides/set-metadata-properties-on-creation)
- Auth0 SDK's [`createUser` documentation](https://auth0.com/docs/libraries/auth0-swift/database-authentication#signing-up-with-a-database-connection)

### Testing

- [x] This change adds integration/UI test coverage

### Checklist

### DIFF
--- a/lib/src/main/java/com/auth0/android/lock/Lock.java
+++ b/lib/src/main/java/com/auth0/android/lock/Lock.java
@@ -492,6 +492,17 @@ public class Lock {
         }
 
         /**
+         * Additional user metadata for Database Sign Up. Empty by default.
+         *
+         * @param userMetadata a non-null Map containing the parameters as Key-Values
+         * @return the current builder instance
+         */
+        public Builder withSignUpUserMetadata(HashMap<String, String> userMetadata) {
+            options.setCustomSignUpUserMetadata(userMetadata);
+            return this;
+        }
+
+        /**
          * Sets the Scope to request when performing the Authentication.
          *
          * @param scope to use in the Authentication.
--- a/lib/src/main/java/com/auth0/android/lock/LockActivity.java
+++ b/lib/src/main/java/com/auth0/android/lock/LockActivity.java
@@ -413,6 +413,8 @@ public class LockActivity extends AppCompatActivity implements ActivityCompat.On
             return;
         }
 
+        event.setExtraFields(options.getCustomSignUpUserMetadata());
+
         AuthenticationAPIClient apiClient = options.getAuthenticationAPIClient();
         final String connection = configuration.getDatabaseConnection().getName();
         lockView.showProgress(true);
--- a/lib/src/main/java/com/auth0/android/lock/events/DatabaseSignUpEvent.java
+++ b/lib/src/main/java/com/auth0/android/lock/events/DatabaseSignUpEvent.java
@@ -44,11 +44,13 @@ public class DatabaseSignUpEvent extends DatabaseEvent {
     @NonNull
     private String password;
     private Map<String, Object> rootAttributes;
+    private Map<String, Object> userMetadata;
 
     public DatabaseSignUpEvent(@NonNull String email, @NonNull String password, @Nullable String username) {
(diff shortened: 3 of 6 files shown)


lbalmaceda:
@morganlutz Hi. How is this PR different from https://github.com/auth0/Lock.Android/pull/543?

morganlutz:
@lbalmaceda Hi! I need to send some additional user metadata in the sign up request that should not be displayed in the Lock UI with a `CustomField`. I do not want Lock UI for this meta data. With this change I can add meta data to the lock builder before launching `LockActivity`.

lbalmaceda:
@morganlutz May I ask what's your use case? What values do you plan to send on those fields?

ejensen:
I put up a corresponding PR for Lock on iOS https://github.com/auth0/Lock.swift/pull/570

morganlutz:
@lbalmaceda If we want to do something like pass in the avatar url, then we'd like to be able to display custom ui for that and pass it ourselves instead of having an extra `CustomField` where the user has to paste in their desired image url. 

I also exposed root attributes so that we can send `app_metadata` like in [this example](https://auth0.com/docs/users/guides/set-metadata-properties-on-creation):

```
{
  "email": "jane.doe@example.com",
  "user_metadata": {
    "hobby": "surfing"
  },
  "app_metadata": {
    "plan": "full"
  }
}
```

lbalmaceda:
@morganlutz There isn't much sense in setting `app_metadata` from a public client such as a mobile app, as that code can easily be decompiled and changed by anyone. I asked the use case because most of the times if you're not asking for explicit user input is because the value you set is already known in advance or can be set later in the sign up flow. e.g. wouldn't it be better to take the default avatar associated to that account (i.e. gravatar) and later in a profile screen allow the user to edit it?

I was going to suggest you use "Auth0 Rules" instead and set the values in there after a sign-up. Alternatively, you could also issue a Management API request and edit the profile yourself. I know however that not all attributes are editable once the user is created, so depending on what you want to set it may or may not work for you.

Have you considered these options already?

ejensen:
@lbalmaceda I added a couple use case examples in https://github.com/auth0/Lock.swift/pull/570#issuecomment-531918692

morganlutz:
@lbalmaceda commit f4ba59f8f91d32febd4bcfc688a59f5c2c510a6a removes the option to assign root attributes

lbalmaceda:
@morganlutz We've reached the following decision: We want to provide this feature as a new type of field: "hidden". This way, you don't need to change any of the public API (e.g. no need to add new methods to the Lock builder). 

`CustomField` already accepts a `hint` and `icon` which are not going to be considered for the hidden type. And creating another constructor just for the hidden type wouldn't feel right because people could still call the current one, pass "hidden" for type expecting it to work OK OOTB, but never ever pass the actual "value" to be sent along with it. I think changes go beyond this `CustomField` class. 

The best solution that fits what we decided as a team is making `CustomField` class extend some interface and change the "withSignUpFields" signature on the builder to accept this interface vs a particular class. On the side of the UI, we would skip displaying them but still iterating through them to fetch the fixed value to send.

I'll be doing this changes for you and close this PR once my PR is published. And I'll work with @cocojoe to provide the analogous steps for the iOS lib. Thanks for raising this and for the time invested so far.

morganlutz:
Thanks @lbalmaceda. Appreciate the quick turnaround 🙌

PR #552, opened 2019-10-07
Add hidden extra sign up fields

lbalmaceda:
### Changes

Previously the extra sign up fields were limited to **visible** fields of type: name, email, phone number and number. This PR extends the support to **hidden** fields that are not being displayed in the sign up form but are still sent with a fixed value to the backend as root profile or user_metadata attributes.

I'm not a fan of keeping reference to `CustomField` type constants from the `HiddenField` class. Ideally these interface definition should be outside or in the parent abstract class. But doing that now would be a breaking change.

#### Usage:
Same builder method as before now accepts subclasses of `SignUpField` class.

##### Previously

```java
List<CustomField> fields = new ArrayList<>();
fields.add(new CustomField(R.drawable.com_auth0_lock_ic_mobile, CustomField.FieldType.TYPE_NAME, "full_name", R.string.hint_name, CustomField.Storage.PROFILE_ROOT));
builder.withSignUpFields(fields);
```

##### Now
Note the method will still work accepting the `List<CustomField>` argument, but if you want to include `HiddenField` and any other future field you'd need to change the list type to `List<SignUpField>`.

```java
List<SignUpField> fields = new ArrayList<>();
fields.add(new CustomField(R.drawable.com_auth0_lock_ic_mobile, CustomField.FieldType.TYPE_NAME, "full_name", R.string.hint_name, CustomField.Storage.PROFILE_ROOT));
fields.add(new HiddenField("reference", "1234", CustomField.Storage.PROFILE_ROOT));
builder.withSignUpFields(fields);
```
### References

- Supersedes https://github.com/auth0/Lock.Android/pull/549

### Testing

### Checklist

### DIFF
--- a/lib/src/main/java/com/auth0/android/lock/utils/HiddenField.java
+++ b/lib/src/main/java/com/auth0/android/lock/utils/HiddenField.java
@@ -1,0 +1,51 @@
+package com.auth0.android.lock.utils;
+
+import android.os.Parcel;
+import android.os.Parcelable;
+import android.support.annotation.NonNull;
+
+public class HiddenField extends SignUpField {
+
+    private final String value;
+
+    /**
+     * Constructor for HiddenField instance
+     *
+     * @param key     the key to store this field as.
+     * @param value   the fixed value to set for this field
+     * @param storage the location to store this value into.
+     */
+    public HiddenField(@NonNull String key, @NonNull String value, @CustomField.Storage int storage) {
+        super(key, storage);
+        this.value = value;
+    }
+
+    @NonNull
+    public String getValue() {
+        return value;
+    }
+
+    protected HiddenField(Parcel in) {
+        super(in);
+        value = in.readString();
+    }
+
+    @Override
+    public void writeToParcel(Parcel dest, int flags) {
+        super.writeToParcel(dest, flags);
+        dest.writeString(value);
+    }
+
+    @SuppressWarnings("unused")
+    public static final Parcelable.Creator<HiddenField> CREATOR = new Parcelable.Creator<HiddenField>() {
+        @Override
+        public HiddenField createFromParcel(Parcel in) {
+            return new HiddenField(in);
+        }
+
+        @Override
+        public HiddenField[] newArray(int size) {
+            return new HiddenField[size];
+        }
+    };
+}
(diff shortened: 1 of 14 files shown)


lib/src/main/java/com/auth0/android/lock/utils/HiddenField.java
```diff
@@ -0,0 +1,51 @@
...
+public class HiddenField extends SignUpField {
+
+    private final String value;
+
+    /**
+     * Constructor for HiddenField instance
+     *
+     * @param key     the key to store this field as.
+     * @param value   the fixed value to set for this field
+     * @param storage the location to store this value into.
```
lbalmaceda: Document the supported "keys" when the storage is "root profile". (or link to docs)

lib/src/main/java/com/auth0/android/lock/utils/SignUpField.java
```diff
@@ -0,0 +1,52 @@
...
+import static com.auth0.android.lock.utils.CustomField.Storage.PROFILE_ROOT;
+
+public abstract class SignUpField implements Parcelable {
+
+    private final String key;
+    private final int storage;
+
+    SignUpField(@NonNull String key, @CustomField.Storage int storage) {
+        if (key.isEmpty()) {
+            throw new IllegalArgumentException("The key cannot be empty!");
```
joshcanhelp:
```suggestion
            throw new IllegalArgumentException("The key cannot be empty.");
```

lbalmaceda: I fear if I change this it would fail if a test is associated. Let me check this locally first.

lbalmaceda: I was right! Resolved on a separate commit.

lib/src/main/java/com/auth0/android/lock/utils/HiddenField.java
lbalmaceda: Resolving because this is already added here https://github.com/auth0/Lock.Android/blob/master/lib/src/main/java/com/auth0/android/lock/utils/CustomField.java#L69

Issue #567, opened 2020-04-28
EditTexts are not showing up


BesatZardoshtCarfax:
### Description

I have completed all the steps from the documentation and I can see the lock activity but there is no input field in this activity.

<img width="322" alt="Screen Shot 2020-04-28 at 5 27 45 AM" src="https://user-images.githubusercontent.com/52752909/80471517-83fbfb80-8911-11ea-8ef2-37d0f9d7ac67.png">

Edward-Kaschula:
I see this in all my environments including production. Downloaded our app from store now to check and there is no username and password fields showing on Login Screen and no fields on the sign up screen either.

Edward-Kaschula:
I found that if I change my build.gradle to 
implementation 'com.google.android.material:material:1.0.0'
rather than the newer 
implementation 'com.google.android.material:material:1.1.0'
everything works again

BesatZardoshtCarfax:
> I found that if I change my build.gradle to
> implementation 'com.google.android.material:material:1.0.0'
> rather than the newer
> implementation 'com.google.android.material:material:1.1.0'
> everything works again

For me it's already 1.1.0

Edward-Kaschula:
Yes so change it to 1.0.0 that will fix the problem

BesatZardoshtCarfax:
Thanks.. That did fix the issue! But I need the latest material library! I still think this issue is valid since the library is not compatible with the latest material library

Edward-Kaschula:
Yes, this needs to be fixed so that we can use the correct Google libraries. The solution here is only a temporary work around. @Auth0 please fix this.

ivabra:
Got the same problem and I've done too far for rolling back to the previous version of material components library, so I put some efforts to get it work for me.

BesatZardoshtCarfax:
Amazing.. Hopefully the review would be quick..

lbalmaceda:
👋 Thanks for creating this. It's indeed an error that needs to be addressed. 

However, I'm trying to reproduce the issue and can't. I've created a new android project from scratch, using Android X and Lock 2.18.0. After following the steps on the Lock's readme, I obtain a simple app that's only meant to launch the LockActivity in order to test the tabs component. Adding the dependency `'com.google.android.material:material:1.1.0'` and even changing it to 1.0.0 doesn't trigger this error. Can someone share more details or even a simple repro project I could try the PR on? For example, AndroidX or AppCompat libraries, dependencies included, API version targetted, etc.

[...]

PR #568, opened 2020-04-28
Fixed material library 1.1.0 tab layout height issue

ivabra:
This pull request refers to #567 . Turned out it is the tab layout that has been stretched vertically so it pushes down the TextViews.

Layout inspector

### Changes

- Layout files of tab layout (`com_auth0_lock_tab.xml`) and tabs (`com_auth0_lock_tab.xml`). The whole idea is to get rid of the strip in the tab's relative layout that has `android:layout_alignParentBottom="true"` which stretches the tab vertically.

#### Screenshots

Before:

After

### References

[Relative Layout docs](https://developer.android.com/reference/android/widget/RelativeLayout.html):
> Note that you cannot have a circular dependency between the size of the RelativeLayout and the position of its children. For example, you cannot have a RelativeLayout whose height is set to `WRAP_CONTENT` and a child set to `ALIGN_PARENT_BOTTOM`.

### Testing

All the automated tests have been passed on a local machine.

The lib needs to be integrated into an application project that uses androidX and  `com.google.android.material:material:1.1.0` dependency to show the real impact of this pull request. It was done locally, no automated tests were added.

### Checklist

### DIFF
--- a/lib/src/main/res/layout/com_auth0_lock_tab.xml
+++ b/lib/src/main/res/layout/com_auth0_lock_tab.xml
@@ -12,9 +12,4 @@
         android:gravity="center"
         android:textAppearance="@style/Lock.Theme.TextAppearance.Tab" />
 
-    <FrameLayout
-        android:layout_width="match_parent"
-        android:layout_height="@dimen/com_auth0_lock_widget_tab_strip_height"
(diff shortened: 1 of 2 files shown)


lbalmaceda:
Thanks @ivabra

Issue #573, opened 2020-05-05
Sign in and Sign up tabs are not changing correctly all the time


BesatZardoshtCarfax:
### Description

When switching between sign in and sign up tabs sometimes the tab does change but not the content of the page.

### Reproduction

- Click on sign in and sign out several times
<img width="318" alt="Screen Shot 2020-05-04 at 9 22 38 PM" src="https://user-images.githubusercontent.com/52752909/81027948-8f897e00-8e4d-11ea-901e-f4bd8eb49b28.png">
<img width="319" alt="Screen Shot 2020-05-04 at 9 23 41 PM" src="https://user-images.githubusercontent.com/52752909/81027951-91534180-8e4d-11ea-87ad-c46aa943257e.png">

### Environment

- implementation 'com.auth0.android:lock:2.19.0'

lbalmaceda:
@BesatZardoshtCarfax Thanks! 

The form displays the right data, as I see the tab title getting its bold style. But the selected tab is not the correct one. 

The logic for updating that view is defined [here](https://github.com/auth0/Lock.Android/blob/master/lib/src/main/java/com/auth0/android/lock/views/ModeSelectionView.java#L84). If this is a sync issue we might need to wrap both selecting the tab and changing the font style around brackets.

In order to define the priority of this, I have a few questions. 
1. How fast are you changing tabs? 
2. Has this started to happen on this last release or is it also present on an old version?
3. Would you mind recording a short video?

amankumar367:
@lbalmaceda ,

I'm also facing the same issue after resolving the issue of Tab height (Material Design version issue)

> How fast are you changing tabs?

Sometimes its also not working in slow tap

> Has this started to happen on this last release or is it also present on an old version?

Yes, After upgrading from `v2.18.0 to v2.19.0`. In the old version, there was no issue with tab changes.

> Would you mind recording a short video?

Please find the below GIF

![Auth0_Lock_Tab_change_issue](https://user-images.githubusercontent.com/35893342/81301093-3a8c7880-9096-11ea-83aa-1c9b8da4ed5c.gif)

lbalmaceda:
Thanks for those details, Aman. I'll add this to our backlog and I'll ask the team if we can prioritize it for the coming week.

PR #577, opened 2020-05-21
Fix ModeSelectionView sync issue

lbalmaceda:
### Changes

Fix a bug on the Mode Selection View, where tabs would appear to be out of sync with the currently selected mode.

This behavior appeared after the layout change in https://github.com/auth0/Lock.Android/pull/568

Basically, before that, the custom tab layout took the whole tab space and the click listener attached to it, was listening for events on the entire tab. 

After the PR got merged, the custom tab layout changed and now only takes 1/3 of the tab height, matching the heigh required by the text that is being displayed. The click listener is still being attached to that custom tab layout, but given it's thinner than before when clicked outside of it, it won't fire any issues. See the image below:

image

### References
Fixes https://github.com/auth0/Lock.Android/issues/573

### Testing

- [ ] This change adds unit test coverage

### Checklist


lib/src/main/java/com/auth0/android/lock/views/ModeSelectionView.java
```diff
@@ -94,19 +96,16 @@ private void toggleBoldText(View tabView, boolean bold) {
         text.setTypeface(bold ? text.getTypeface() : null, bold ? Typeface.BOLD : Typeface.NORMAL);
     }
 
-    @Deprecated
```
jimmyjames, line 97: Why are removing the deprecation of these methods?

lbalmaceda: I don't know why they were added, to begin with. The methods are not deprecated on the android SDK, otherwise, I'd receive a warning on the IDE (see screenshot). 

I assume they were marked as deprecated since they do nothing and are no longer used (see no-op). But as any "custom view" we include on this SDK, for me those are considered **internal** and shouldn't be used directly. I can't imagine someone adding a piece of our form to their app if it just won't work OOTB without the rest of the classes, for example. But still, they are public and anyone who could be calling those methods could find it breaking if we remove them.

image


@jimmyjames do you think I should keep the deprecated annotation? I'm OK with that as well.

jimmyjames: Thanks for the explanation @lbalmaceda. I think the change to remove the deprecation is fine per your reasoning 👍

PR #589, opened 2020-08-21
Support bot protection

lbalmaceda:
### Changes

When a request is flagged as suspicious by the Bot Protection feature, it will require the user to complete the authentication on a web-based flow. When this error code happens on Lock, the browser will be automatically launched on the right page (login/signup).

### References

https://auth0.com/docs/anomaly-detection/bot-protection
### Testing

Added unit tests for the event handling. I've also manually tested sign in, sign up and log in.

### Checklist


lib/build.gradle
```diff
@@ -36,6 +36,10 @@ android {
     testOptions {
         unitTests {
             includeAndroidResources = true
+            // https://github.com/robolectric/robolectric/issues/5115
+            all {
+                systemProperty("javax.net.ssl.trustStoreType", "JKS")
```
lbalmaceda, line 41: Required to be set before `MockServer` can be instantiated. See the referenced issue for details.

lib/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -120,6 +122,9 @@ public LockActivity() {
         this.options = options;
         this.lockView = lockView;
         this.webProvider = webProvider;
+        this.handler = new Handler();
```
lbalmaceda, line 125: These are only used in tests. In absolutely NO scenario Android will instantiate the LockActivity through this constructor, so it's safe to use.

lib/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -372,6 +377,27 @@ public void onOAuthAuthenticationRequest(OAuthLoginEvent event) {
         webProvider.start(this, connection, extraAuthParameters, authProviderCallback, WEB_AUTH_REQUEST_CODE);
     }
 
+    private void completeDatabaseAuthenticationOnBrowser() {
+        //DBConnection checked for nullability before the API call
```
lbalmaceda, line 381: This is called from a `callback#onFailure`. The action that registered the callback already checked for nullability, see https://github.com/auth0/Lock.Android/pull/589/files#diff-01fe7e862df1aaab8f18b1b31542c58fR404-R406.

lib/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -372,6 +377,27 @@ public void onOAuthAuthenticationRequest(OAuthLoginEvent event) {
...
+        @SuppressWarnings("ConstantConditions")
+        String connection = configuration.getDatabaseConnection().getName();
+
+        String loginHint = null;
+        String screenHint = null;
+        if (lastDatabaseSignUp != null) {
+            loginHint = lastDatabaseSignUp.getEmail();
+            screenHint = "signup";
```
lbalmaceda, line 389: thought of using constants for these, but they are only being used here, once.

lib/src/test/java/com/auth0/android/lock/LockActivityTest.java
```diff
@@ -203,6 +209,124 @@ public void shouldCallOIDCDatabaseLoginWithOTPCodeAndMFAToken() {
...
+        Auth0 account = new Auth0("cliendId", "domain");
+        account.setOIDCConformant(true);
+        Options options = mock(Options.class);
+        when(options.getAccount()).thenReturn(account);
+        when(options.getAuthenticationAPIClient()).thenReturn(client);
+        when(options.getScope()).thenReturn("openid user photos");
+        when(options.getAudience()).thenReturn("aud");
+        when(options.getAuthenticationParameters()).thenReturn(basicParameters);
+        //set sign in behavior
+        when(options.loginAfterSignUp()).thenReturn(true);
```
lbalmaceda, line 262: this is the important line

lib/src/test/java/com/auth0/android/lock/LockActivityTest.java
```diff
@@ -203,6 +209,124 @@ public void shouldCallOIDCDatabaseLoginWithOTPCodeAndMFAToken() {
...
+        Auth0 account = new Auth0("cliendId", "domain");
+        account.setOIDCConformant(true);
+        Options options = mock(Options.class);
+        when(options.getAccount()).thenReturn(account);
+        when(options.getAuthenticationAPIClient()).thenReturn(client);
+        when(options.getScope()).thenReturn("openid user photos");
+        when(options.getAudience()).thenReturn("aud");
+        when(options.getAuthenticationParameters()).thenReturn(basicParameters);
+        //set create user behavior
+        when(options.loginAfterSignUp()).thenReturn(false);
```
lbalmaceda, line 304: this is the important line

lib/src/test/java/com/auth0/android/lock/LockActivityTest.java
```diff
@@ -416,7 +540,7 @@ public void shouldCallOAuthAuthenticationWithCustomProvider() {
         verify(lockView, never()).showProgress(true);
         verify(customProvider).setParameters(mapCaptor.capture());
         verify(customProvider).start(eq(activity), any(AuthCallback.class), eq(REQ_CODE_PERMISSIONS), eq(REQ_CODE_CUSTOM_PROVIDER));
-        AuthResolver.setAuthHandlers(Collections.emptyList());
+        AuthResolver.setAuthHandlers(Collections.<AuthHandler>emptyList());
```
lbalmaceda, line 543: these were required after fixing the wrong class import

lib/src/test/java/com/auth0/android/lock/PasswordlessLockActivityTest.java
```diff
@@ -33,11 +33,10 @@
 import org.robolectric.RobolectricTestRunner;
 import org.robolectric.annotation.Config;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
-import edu.emory.mathcs.backport.java.util.Collections;
```
lbalmaceda, line 39: incorrect class, omg.

lib/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -552,11 +584,16 @@ public void run() {
...
                 }
             });
+            lastDatabaseSignUp = null;
         }
 
         @Override
         public void onFailure(final AuthenticationException error) {
             Log.e(TAG, "Failed to create the user: " + error.getMessage(), error);
+            if (error.isVerificationRequired()) {
+                completeDatabaseAuthenticationOnBrowser();
```
Widcket, line 594: Why on this error callback the `completeDatabaseAuthenticationOnBrowser()` method is called before (outside) the runnable, and in the other error handler (in `authCallback`) it's called inside the runnable?

lbalmaceda: It's actually better to move it outside of the runnable. 👍 

The reason that runnable exists is that the UI is touched on the callback, and it needs to be touched in the main/UI thread. The UI was being updated to hide the progress bar, but it actually makes sense to keep showing it until it comes back from the browser (either with a good callback or because the user canceled the auth). I'll make this change.

.circleci/config.yml
```diff
@@ -2,23 +2,20 @@ version: 2
...
     - save_cache:
         key: v1-dep-{{ .Branch }}-{{ epoch }}
         paths:
         - ~/.gradle
         - ~/.android
         - /usr/local/android-sdk-linux/extras
-    
-    - run: ./gradlew clean test --continue --console=plain
\ No newline at end of file
+    - run: ./gradlew clean test --continue --console=plain --max-workers=4
```
lbalmaceda, line 21: limited the number of workers to avoid OOM issues

lib/build.gradle
```diff
@@ -36,6 +36,11 @@ android {
...
+                maxHeapSize = "1024m"
```
lbalmaceda, line 42: these are independent of the JVM options passed in the command line (see CI script)

lib/src/main/java/com/auth0/android/lock/LockActivity.java
lbalmaceda: Moved out of the runnable

PR #610, opened 2021-04-13
Use Auth0.Android v2 [SDK-2429]

lbalmaceda:
### Changes

Targetting the newer Auth0 SDK imposes two new constraints/requirements:
- This SDK will now target SDK 21 and above (#608)
- This SDK will now require to be run in a JDK 8 target

Once these settings have been adjusted, the [migration guide](https://github.com/auth0/Auth0.Android/blob/main/V2_MIGRATION_GUIDE.md) was followed to fix the compiling errors. These mostly were:
- Authentication parameters type `Map<String, Object>` --> `Map<String, String>`
- Callback types `AuthCallback` --> `Callback`
- Removed classes: `VoidCallback` and `Telemetry`
- Removed methods: `OIDC conformant`

**Note:** The `Auth0Parcelable` class, used to communicate the `Auth0` class settings between Lock instances, is not completely migrated. As of this PR, it's lacking support for customizing the networking client; a feature that was recently made available in Auth0 Android v2. A comment was added to re-evaluate this in the short term.

### References

See `SDK-2429`.

### Testing

Compile errors in the tests have not been completely fixed yet (wip). Once they are fixed, this PR will be removed from DRAFT.


lib/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -372,12 +374,12 @@ public void onOAuthAuthenticationRequest(@NonNull OAuthLoginEvent event) {
...
 
-        Map<String, Object> extraAuthParameters = null;
+        Map<String, String> extraAuthParameters = null;
         if (!TextUtils.isEmpty(event.getUsername())) {
-            extraAuthParameters = Collections.singletonMap(KEY_LOGIN_HINT, (Object) event.getUsername());
+            extraAuthParameters = Collections.singletonMap(KEY_LOGIN_HINT, event.getUsername());
         }
         Log.d(TAG, "Couldn't find an specific provider, using the default: " + WebAuthProvider.class.getSimpleName());
-        webProvider.start(this, connection, extraAuthParameters, authProviderCallback, WEB_AUTH_REQUEST_CODE);
+        webProvider.start(this, connection, extraAuthParameters, new WebCallbackWrapper(authProviderCallback));
```
lbalmaceda, line 382: Lock works with the concept of `AuthProvider`. These are implementations from users (or our library) that know different strategies to sign in. We provide one for the web auth flow, using our `WebAuthProvider`. 

Because the `AuthProvider` interface takes an `AuthCallback` to return the results, and given that we've changed the callback type in the last SDK release, we have to wrap it in this new class `WebCallbackWrapper` (internal)

lib/src/main/java/com/auth0/android/lock/WebProvider.java
```diff
@@ -88,17 +87,4 @@ public boolean resume(Intent intent) {
...
 
-    /**
-     * Finishes the authentication flow in the WebAuthProvider
-     *
-     * @param requestCode the request code received on the onActivityResult method
-     * @param resultCode  the result code received on the onActivityResult method
-     * @param intent      the intent received in the onActivityResult method.
-     * @return true if a result was expected and has a valid format, or false if not
-     */
-    public boolean resume(int requestCode, int resultCode, Intent intent) {
```
lbalmaceda, line 99: Deprecated, removed. This class is internal anyway.

lib/src/main/java/com/auth0/android/lock/internal/configuration/ApplicationFetcher.java
```diff
@@ -25,122 +25,131 @@
...
-        Uri uri = Uri.parse(account.getConfigurationUrl()).buildUpon().appendPath("client")
-                .appendPath(account.getClientId() + ".js").build();
+    private static class FetchTask extends AsyncTask<Callback<List<Connection>, Auth0Exception>, Void, List<Connection>> {
 
-        Request req = new Request.Builder()
-                .url(uri.toString())
-                .build();
+        private final Auth0 account;
 
-        client.newCall(req).enqueue(new Callback() {
```
lbalmaceda, line 87: We will no longer use the OkHttp client directly, but through the interface exposed by the `Auth0` class. We still have to do the error handling, as we're doing here.

The logic in this class has been moved to a "background task" `AsyncTask` where the network request will be safely executed.

lib/src/main/java/com/auth0/android/lock/internal/configuration/Connection.java
```diff
@@ -22,9 +21,13 @@
     private PasswordComplexity passwordComplexity;
 
     private Connection(@NonNull String strategy, Map<String, Object> values) {
-        checkArgument(values != null && values.size() > 0, "Must have at least one value");
```
lbalmaceda, line 25: This method went away. We're doing it here now instead.

build.gradle
```diff
@@ -17,6 +17,7 @@ allprojects {
     group = 'com.auth0.android'
     
     repositories {
+        mavenLocal()
```
lbalmaceda, line 20: temporary until we make a new release of the sdk

lib/src/main/java/com/auth0/android/lock/events/DatabaseSignUpEvent.java
```diff
@@ -66,19 +68,14 @@ public void setRootAttributes(@NonNull Map<String, Object> attributes) {
...
     @SuppressWarnings("ConstantConditions")
     @NonNull
     public SignUpRequest getSignUpRequest(@NonNull AuthenticationAPIClient apiClient, @NonNull String connection) {
-        SignUpRequest request;
-        if (getUsername() != null) {
-            request = apiClient.signUp(getEmail(), getPassword(), getUsername(), connection);
-        } else {
-            request = apiClient.signUp(getEmail(), getPassword(), connection);
-        }
+        SignUpRequest request = apiClient.signUp(getEmail(), getPassword(), getUsername(), connection, userMetadata);
```
lbalmaceda, line 78: user metadata will be passed as an extra param of this method call, rather than later in the request parameters. The method accepts a nullable username, so we no longer need the if

lib/src/test/java/com/auth0/android/lock/WebProviderTest.java
```diff
@@ -202,55 +201,9 @@ public void shouldStartBrowserWithOptions() {
...
+        assertThat(authorizeUri, hasParamWithValue("scope", "email profile photos openid"));
         assertThat(authorizeUri, hasParamWithValue("connection_scope", "the connection scope"));
-        assertThat(intent.hasExtra("com.auth0.android.EXTRA_USE_BROWSER"), is(true));
         assertThat(intent.getParcelableExtra("com.auth0.android.EXTRA_CT_OPTIONS"), is(notNullValue()));
-        assertThat(intent.getBooleanExtra("com.auth0.android.EXTRA_USE_BROWSER", false), is(true));
-        assertThat(intent, hasComponent(AuthenticationActivity.class.getName()));
-    }
-
-    @Test
-    public void shouldStartWebViewWithOptions() {
```
lbalmaceda, line 214: "webview" (instead of an external browser) is no longer supported

lib/src/test/java/com/auth0/android/lock/WebProviderTest.java
```diff
@@ -262,14 +215,6 @@ public void shouldResumeWithIntent() {
         assertThat(webProvider.resume(intent), is(false));
     }
 
-    @Test
-    public void shouldResumeWithCodesAndIntent() {
-        Intent intent = mock(Intent.class);
-        Options options = mock(Options.class);
-        WebProvider webProvider = new WebProvider(options);
-        assertThat(webProvider.resume(1, Activity.RESULT_CANCELED, intent), is(false));
```
lbalmaceda, line 270: launching with a result code is no longer valid. This was only used for the "webview" flow.

lib/src/test/java/com/auth0/android/lock/errors/SignUpErrorMessageBuilderTest.java
```diff
@@ -27,7 +27,7 @@
 
     @Before
     public void setUp() {
-        MockitoAnnotations.initMocks(this);
+        MockitoAnnotations.openMocks(this);
```
lbalmaceda, line 30: the previous one is marked as deprecated. In some classes, I had to change it to this new way for the tests to run successfully.

lib/src/test/java/com/auth0/android/lock/utils/ApplicationAPI.java
```diff
@@ -35,11 +35,8 @@
 
     private MockWebServer server;
 
-    public ApplicationAPI() throws IOException {
-        // FIXME: Remove this if migrating to Robolectric 4.4
```
lbalmaceda, line 39: this PR migrates to robolectric 4.4; And this code is no longer needed.

lib/src/test/java/com/auth0/android/lock/utils/SSLTestUtils.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Collections;
+
+import okhttp3.mockwebserver.MockWebServer;
+import okhttp3.tls.HandshakeCertificates;
+import okhttp3.tls.HeldCertificate;
+
+/**
+ * Utility object for executing tests that use the networking client over HTTPS on localhost.
+ */
+public class SSLTestUtils {
```
lbalmaceda, line 16: this is a test helper class. Not included as part of the packaged library

lib/src/test/java/com/auth0/android/lock/LockActivityTest.java
```diff
@@ -106,13 +118,15 @@ public void setUp() {
         when(client.login(anyString(), anyString(), anyString())).thenReturn(authRequest);
         when(client.loginWithOTP(anyString(), anyString())).thenReturn(authRequest);
         when(client.createUser(anyString(), anyString(), anyString())).thenReturn(dbRequest);
-        when(client.createUser(anyString(), anyString(), anyString(), anyString())).thenReturn(dbRequest);
+        when(client.createUser(anyString(), anyString(), any(), anyString())).thenReturn(dbRequest);
```
lbalmaceda, line 121: `any()` required for the username parameter, as it could be null or a valid string

lib/build.gradle
```diff
@@ -58,14 +63,15 @@ dependencies {
     implementation 'androidx.legacy:legacy-support-v4:1.0.0'
     implementation 'com.google.android.material:material:1.3.0'
     implementation 'com.google.code.gson:gson:2.8.6'
-    implementation 'com.squareup.okhttp:okhttp:2.7.5'
     implementation 'com.squareup:otto:1.3.8'
-    api 'com.auth0.android:auth0:1.29.0'
+    api 'com.auth0.android:auth0:2.1.0-SNAPSHOT'
```
jimmyjames, line 67: also temporarily using a snapshot build built locally; makes sense for now

lib/src/main/java/com/auth0/android/lock/Auth0Parcelable.java
```diff
@@ -61,14 +62,10 @@ public void writeToParcel(@NonNull Parcel dest, int flags) {
...
         dest.writeString(auth0.getConfigurationUrl());
-        dest.writeByte((byte) (auth0.isOIDCConformant() ? WITH_DATA : WITHOUT_DATA));
-        dest.writeByte((byte) (auth0.isLoggingEnabled() ? WITH_DATA : WITHOUT_DATA));
-        dest.writeByte((byte) (auth0.getTelemetry() != null ? WITH_DATA : WITHOUT_DATA));
-        if (auth0.getTelemetry() != null) {
-            dest.writeString(auth0.getTelemetry().getName());
-            dest.writeString(auth0.getTelemetry().getVersion());
-            dest.writeString(auth0.getTelemetry().getLibraryVersion());
-        }
+        //FIXME: Find a way to pass the NetworkingClient implementation
```
jimmyjames, line 65: Mostly just for my own education, but what does this comment mean?

jimmyjames (commented):
Left a few minor comments; may be good to walk through the changes together as it gets closer to being out of draft.

lib/src/main/java/com/auth0/android/lock/Auth0Parcelable.java
lbalmaceda: This relates to the "note" I added to the PR description. Customizing the networking client was made available in the latest release of the SDK. But the way we accept that client is with a setter in the `Auth0` class. 

This library has an implementation that makes `Auth0` "Parcelable" (similar to `Serializable`), so we could pass it across instances/classes to set up the widget. But because the `NetworkingClient` implementation cannot be serialized, it's not straightforward to update the `Auth0` class to support it. I added the comment to come back to this later and evaluate if this is something we need to backport today or not.

jimmyjames (approved):

lib/src/main/java/com/auth0/android/lock/Auth0Parcelable.java
NicholasGWK: Hello there! I realize this is a _very_ old PR, but I was wondering if this was ever floated to be fixed, as I noticed it's still in the current version...being able to pass in a custom networking client would be super useful for my team right now as we have to do weird proxy stuff but would prefer not to do a full migration to the raw SDK at the moment 😅 

My understanding is lock is not really receiving any new feature requests, but wondering if this would qualify as a bugfix and if we can contribute to a PR somehow!

Thanks so much!

PR #616, opened 2021-04-21
Migrate sample app to use Kotlin [SDK-2431]

lbalmaceda:
### Changes

This is a small refactor improvement anticipating the new README changes, that will have snippets showing Kotlin usage. The code was generated using the java->kotlin code conversion assistant, with manual amendments that followed.

### References

See SDK-2431

### Testing

I've manually ran the app and tested the typical workflows


app/src/main/java/com/auth0/android/lock/app/DemoActivity.kt
```diff
@@ -0,0 +1,260 @@
...
+import com.auth0.android.Auth0
+import com.auth0.android.authentication.AuthenticationException
+import com.auth0.android.callback.Callback
+import com.auth0.android.lock.*
+import com.auth0.android.lock.utils.LockException
+import com.auth0.android.provider.WebAuthProvider.login
+import com.auth0.android.provider.WebAuthProvider.logout
+import com.auth0.android.result.Credentials
+import com.google.android.material.snackbar.Snackbar
+import java.util.*
```
Widcket, line 42: Why not import just the utils you need, like in the Java file?

Widcket (dismissed):
LGTM, just have a question

app/src/main/java/com/auth0/android/lock/app/DemoActivity.kt
lbalmaceda: `java.util.*` was, in fact, an unused import. I've removed it. I've also tried to simplify the `lock.*` one but the IDE preferred that one after importing a few individual classes and re-organizing the imports.

Widcket (approved):

PR #617, opened 2021-04-26
Update readme

lbalmaceda:
### Changes
This PR updates the readme using kotlin snippets and removing instructions that no longer apply (e.g. OIDC mode).

The content is not final and might be updated in the following PRs.

### References
See `SDK-2431`

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,14 +123,15 @@ class MyActivity : AppCompatActivity() {
     super.onCreate(savedInstanceState)
     
     val account = Auth0(this)
+    // Instantiate Lock once
     lock = Lock.newBuilder(account, callback)
       // Customize Lock
-      // .withScheme("myapp")
       .build(this)
   }
 
   override fun onDestroy() {
       super.onDestroy()
+      // Important! Release Lock and its resources
       lock?.onDestroy(this)
   }
 
@@ -218,14 +219,15 @@ class MyActivity : AppCompatActivity() {
     super.onCreate(savedInstanceState)
     
     val account = Auth0(this)
+    // Instantiate Lock once
     lock = PasswordlessLock.newBuilder(account, callback)
-      // Customize Lock
-      // .withScheme("myapp")
-      .build(this)
+        // Customize Lock
+        .build(this)
   }
 
   override fun onDestroy() {
       super.onDestroy()
+      // Important! Release Lock and its resources
       lock?.onDestroy(this)
   }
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,33 +70,22 @@ android {
...
 ```
 
-The next step is to create an instance of `Auth0` with your application's information:
-
-```java
-Auth0 account = new Auth0("{YOUR_AUTH0_CLIENT_ID}", "{YOUR_AUTH0_DOMAIN}");
-```
-
-Alternatively, if you are using _custom domains_ and are required to specify a different URL to fetch the Lock widget configuration from, you can use the 3 String constructor:
+The next step is to create an instance of `Auth0` with your application's information. The easiest way to create it is by using the values defined previously in the `string.xml` file and passing an Android Context. For this to work, you must have defined the string resources using the same keys as listed above.
```
Widcket, line 73:
```suggestion
The next step is to create an instance of `Auth0` with your application's information. The easiest way to create it is by using the values defined previously in the `strings.xml` file and passing an Android Context. For this to work, you must have defined the string resources using the same keys as listed above.
```

README.md
```diff
@@ -150,91 +108,57 @@ Next, add the `LockActivity` inside the `application` tag:
...
-        android:pathPrefix="/android/${applicationId}/callback"
-        android:scheme="https" />
-    </intent-filter>
-</activity>
-```
+Make sure the Activity's `launchMode` is declared as `singleTask` or the authentication result won't come back into your application.
 
 
-Make sure the Activity's `launchMode` is declared as `singleTask` or the result won't come back in the authentication.
+Then, in any of your Activities, you need to initialize **Lock** and handle the release of its resources appropriately after you're doing using it.
```
Widcket, line 114:
```suggestion
Then, in any of your Activities, you need to initialize **Lock** and handle the release of its resources appropriately after you're done using it. 
```

Widcket: It's also missing the scheme part: If your `auth0Scheme` value is not `https` you need to use `withScheme()` as shown below.

README.md
```diff
@@ -150,91 +108,57 @@ Next, add the `LockActivity` inside the `application` tag:
...
+      // .withScheme("myapp")
+      .build(this)
   }
 
-  @Override
-  protected void onDestroy() {
-    super.onDestroy();
-    // Your own Activity code
-    lock.onDestroy(this);
-    lock = null;
```
Widcket, line 207: Why is there no need to set the reference to null in the Kotlin version?

Widcket (changes requested):
There are a couple of typos and questions. Other than that, it looks good.

README.md
lbalmaceda: 😮 good catch!!

README.md
lbalmaceda: I'll add a separate section for the main options that can be changed. I wanted to take the scheme one out of the main instructions.

README.md
lbalmaceda: Is not because of kotlin. When an android activity gets its onDestroy method called is because is going away for good. There's no need to set to null the references. The garbage collector will recycle the whole activity instance.

lbalmaceda:
Comments addressed!

Widcket (approved):

PR #618, opened 2021-04-27
Raise Unauthorized and Access Denied errors through callback [SDK-2480]

lbalmaceda:
### Changes

These two error codes `unauthorized` and `access_denied` can be raised through [Auth0 Rules](https://auth0.com/docs/rules). 

To improve the use case of handling custom errors in the authentication pipeline, these will now be raised directly through the provided callback instead of displaying an orange toast / snackbar with a generic error message. That means developers will need to handle this use case from now on. 

**This represents a breaking change in behavior. The public API remains the same after this PR.**

Affected flows: Any authentication flow. (e.g. Passwordless, Database, Enterprise, WebAuth)

### References
See `SDK-2480`

### Testing

There are currently no tests for this part of the widget. This change was tested manually.

### DIFF
--- a/lib/src/main/java/com/auth0/android/lock/Lock.java
+++ b/lib/src/main/java/com/auth0/android/lock/Lock.java
@@ -36,6 +36,7 @@ import androidx.annotation.StyleRes;
 import androidx.localbroadcastmanager.content.LocalBroadcastManager;
 
 import com.auth0.android.Auth0;
+import com.auth0.android.authentication.AuthenticationException;
 import com.auth0.android.lock.LockCallback.LockEvent;
 import com.auth0.android.lock.internal.configuration.Options;
 import com.auth0.android.lock.internal.configuration.Theme;
@@ -150,8 +151,8 @@ public class Lock {
         switch (action) {
             case Constants.AUTHENTICATION_ACTION:
                 Log.v(TAG, "AUTHENTICATION action received in our BroadcastReceiver");
-                if (data.hasExtra(Constants.ERROR_EXTRA)) {
-                    callback.onError(new LockException(data.getStringExtra(Constants.ERROR_EXTRA)));
+                if (data.hasExtra(Constants.EXCEPTION_EXTRA)) {
+                    callback.onError(new LockException((AuthenticationException) data.getSerializableExtra(Constants.EXCEPTION_EXTRA)));
                 } else {
                     callback.onEvent(LockEvent.AUTHENTICATION, data);
                 }
--- a/lib/src/main/java/com/auth0/android/lock/LockActivity.java
+++ b/lib/src/main/java/com/auth0/android/lock/LockActivity.java
@@ -220,6 +220,14 @@ public class LockActivity extends AppCompatActivity implements ActivityCompat.On
         finish();
     }
 
+    private void deliverAuthenticationError(AuthenticationException exception) {
+        Intent intent = new Intent(Constants.AUTHENTICATION_ACTION);
+        intent.putExtra(Constants.EXCEPTION_EXTRA, exception);
+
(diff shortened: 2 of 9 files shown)


lib/src/main/java/com/auth0/android/lock/Lock.java
```diff
@@ -150,8 +151,8 @@ private void processEvent(@NonNull Context context, @NonNull Intent data) {
         switch (action) {
             case Constants.AUTHENTICATION_ACTION:
                 Log.v(TAG, "AUTHENTICATION action received in our BroadcastReceiver");
-                if (data.hasExtra(Constants.ERROR_EXTRA)) {
-                    callback.onError(new LockException(data.getStringExtra(Constants.ERROR_EXTRA)));
+                if (data.hasExtra(Constants.EXCEPTION_EXTRA)) {
```
lbalmaceda, line 154: The former was not used from the code. Replaced with a more meaningful extra that can hold the Exception instance with all the details.

lib/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -506,12 +518,16 @@ public void onSuccess(@Nullable Credentials credentials) {
 
         @Override
         public void onFailure(@NonNull final AuthenticationException error) {
-            Log.e(TAG, "Failed to authenticate the user: " + error.getMessage(), error);
-            final AuthenticationError authError = loginErrorBuilder.buildFrom(error);
+            Log.e(TAG, "Failed to authenticate the user: " + error.getCode(), error);
```
lbalmaceda, line 521: covers the database/enterprise flow for Lock

lib/src/test/java/com/auth0/android/lock/AuthenticationCallbackTest.java
```diff
@@ -80,6 +82,16 @@ public void shouldReturnAuthentication() {
...
     }
 
+    @Test
+    public void shouldReturnAuthenticationError() {
+        Intent data = new Intent();
+        AuthenticationException error = new AuthenticationException("err_code", "err description");
+        data.putExtra(Constants.EXCEPTION_EXTRA, error);
+        callback.onEvent(LockEvent.AUTHENTICATION, data);
+
+        assertThat(callback, hasError());
```
lbalmaceda, line 92: checks that when an "exception extra" is present, the `onError` is invoked with a non-null value for the exception

lib/src/main/java/com/auth0/android/lock/LockActivity.java
```diff
@@ -542,7 +558,7 @@ public void onSuccess(@Nullable final DatabaseUser user) {
 
         @Override
         public void onFailure(@NonNull final AuthenticationException error) {
-            Log.e(TAG, "Failed to create the user: " + error.getMessage(), error);
+            Log.e(TAG, "Failed to create the user: " + error.getCode(), error);
```
lbalmaceda, line 561: Changed a few log lines, moved them up in the code and used the error code instead. These might be removed before the release.

lib/src/main/java/com/auth0/android/lock/Lock.java
```diff
@@ -150,8 +151,8 @@ private void processEvent(@NonNull Context context, @NonNull Intent data) {
...
+                    callback.onError(new LockException((AuthenticationException) data.getSerializableExtra(Constants.EXCEPTION_EXTRA)));
```
Widcket, line 155: Is the deserialization into an `AuthenticationException` expected to always succeed?

Widcket: nvm, I saw below that the AuthenticationException is sent by the library.

Widcket (approved):

Issue #607, opened 2021-02-15
AuthenticationCallback methods are only called the first time I login


DaireJN:
### Describe the problem

the first time i login callbacks are recieved as expected callbacks are not recieved in subsequent logins, if i close the app and attempt to login again the same occurs (first time = callbacks recieved, nothing subsequent times) The logs in the auth0 dashboard indicate everything is ok

### What was the expected behavior?

callback recieved everytime i login or cancel/error

### Sample

Here is a project demontrating the issue, to build it replace "com_auth0_domain" and  "com_auth0_client_id" in strings.xml with valid credentials and run.

https://github.com/DaireJN/LockExample.git

## Logs

these logs are recieved on fist sign in, no logs occur for subsequent sign ins until app is killed + reopened

2021-02-15 17:42:14.646 26962-26962/packagename V/Lock: AUTHENTICATION action received in our BroadcastReceiver
2021-02-15 17:42:14.646 26962-26962/packagename D/AuthenticationCallback: User authenticated!

https://user-images.githubusercontent.com/28805987/107978323-799c1280-6fb4-11eb-9fac-6a28c558036d.mov

### Environment

- **Version of this library used:** 2.24.0
- **Which framework are you using, if applicable:** 
- **Other modules/plugins/libraries that might be involved:** Kotlin
- **Any other relevant information you think would be useful:** min sdk 21 target sdk 29

DaireJN:
Any insights on this?

lbalmaceda:
@DaireJN  Sorry for the delayed response. Thanks for providing a repro project, I was able to reproduce this. 

It seems this [offending line](https://github.com/auth0/Lock.Android/blob/master/lib/src/main/java/com/auth0/android/lock/Lock.java#L150) in the library code is unregistering the even listener after an event was received. As long as devs are calling `lock.onDestroy()` to clear that broadcast listener registration when they decide, we should be able to remove that line safely. 

So TLDR, we have to:
- Remove the line I linked above in both `PasswordlessLock` and `Lock` files.
- Improve the docs around calling `onDestroy` to ensure a proper lifecycle callback handling.

In the meantime, and since the listener is registered every time you `build` a lock instance from the Builder, you can use that to reset the callback registration.

lbalmaceda:
[...]

PR #619, opened 2021-04-27
Fix broadcast receiver usage

lbalmaceda:
### Changes
The broadcast receiver that handles events was being registered right after the lock widget `Builder` was built. This is because the lock instance is meant to be created once and reused. And while this instance is meant to be released in the (dev's) `activity#onDestroy()` method, this wasn't consistently documented across the readme.

Calling multiple times `builder#build(context)` like we used to do in our demo app before this PR, would cause multiple different broadcast receiver instances to be registered. Then the callback would be invoked multiple times after each event. 

The right way to clear up resources and broadcast receiver registrations is via the `lock#onDestroy(context)`.

### References

Fixes https://github.com/auth0/Lock.Android/issues/607

### Testing
This part of the code is not unit-tested.

### DIFF
--- a/lib/src/main/java/com/auth0/android/lock/Lock.java
+++ b/lib/src/main/java/com/auth0/android/lock/Lock.java
@@ -137,12 +137,14 @@ public class Lock {
     }
 
     private void initialize(@NonNull Context context) {
+        LocalBroadcastManager lbm = LocalBroadcastManager.getInstance(context);
+        lbm.unregisterReceiver(this.receiver);
         IntentFilter filter = new IntentFilter();
         filter.addAction(Constants.AUTHENTICATION_ACTION);
         filter.addAction(Constants.SIGN_UP_ACTION);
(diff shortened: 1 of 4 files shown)


lib/src/main/java/com/auth0/android/lock/Lock.java
```diff
@@ -136,16 +136,17 @@ public void onDestroy(@NonNull Context context) {
     }
 
     private void initialize(@NonNull Context context) {
+        LocalBroadcastManager lbm = LocalBroadcastManager.getInstance(context);
+        lbm.unregisterReceiver(this.receiver);
```
lbalmaceda, line 140: this is a no-op when the receiver was not previously registered. Is a safe way of making sure we only register this one once, avoiding the same event to be delivered multiple times to the dev provided callback.

Widcket (approved):

Widcket (approved):

Widcket (approved):

PR #620, opened 2021-04-29
Simplify library set up by declaring activities internally

lbalmaceda:
### Changes
Before this PR, when a developer wanted to use this library, they would need to declare the Lock activities in the `AndroidManifest.xml` file. 

Following what the SDK does, this PR will declare both Classic and Passwordless lock required activities, which will be merged in the final `AndroidManifest.xml` file. This also applies to the "Internet" permission.

This change effectively reduces the changes required to use the library, removing the need to modify the manifest file. 

**This is not a breaking change**, but developers are encouraged to remove the activities declaration from their manifest files to reduce the possibility of duplicated intent filter declarations (no harm).

### References
None

### Testing
Manually tested


stevehobbsdev (dismissed):

lbalmaceda:
@stevehobbsdev had to fix the tests! ready to review again

stevehobbsdev (approved):

PR #621, opened 2021-04-29
Update LockCallback and AuthenticationCallback [SDK-2480]

lbalmaceda:
### Changes

With this change, the `AuthenticationCallback` is reduced from 3 methods down to 2. Before, the "canceled" event would trigger a `onCanceled` method call. This method no longer exists and instead, an exception now is raised with the code "a0.authentication_canceled". This is easily testable checking the `AuthenticationException#isCanceled()` method.

The immediate benefit is accessing the `AuthenticationException` directly, without having to cast the cause wrapped in the (previously existing) `LockException`. This helps to consume different error codes and messages that are now going to be received after https://github.com/auth0/Lock.Android/pull/618

```kotlin
// Before
val callback: LockCallback = object : AuthenticationCallback() {
    override fun onAuthentication(credentials: Credentials) {
        // Authenticated
    }
    override fun onCanceled() {
        // Canceled
    }
    override fun onError(error: LockException) {
        // Another error. Check code & description.
    }
}
// After
val callback: LockCallback = object : AuthenticationCallback() {
    override fun onAuthentication(credentials: Credentials) {
        // Authenticated
    }
    override fun onError(error: AuthenticationException) {
        if (error.isCanceled) {
            // Canceled
        } else {
            // Another error. Check code & description.
        }
    }
}
```

**This PR includes a breaking change.**

### References

See SDK-2480


lib/src/main/java/com/auth0/android/lock/AuthenticationCallback.java
```diff
@@ -42,32 +40,24 @@
...
-                onCanceled();
-                break;
-            case LockEvent.AUTHENTICATION:
-                parseAuthentication(data);
-                break;
-            case LockEvent.RESET_PASSWORD:
-            case LockEvent.SIGN_UP:
-                break;
+        if (data.hasExtra(Constants.EXCEPTION_EXTRA)) {
+            onError((AuthenticationException) data.getSerializableExtra(Constants.EXCEPTION_EXTRA));
```
lbalmaceda, line 53: any exception available would be raised through `onError` immediately

lib/src/main/java/com/auth0/android/lock/AuthenticationCallback.java
```diff
@@ -42,32 +40,24 @@
...
+            return;
+        }
+        if (event == LockEvent.AUTHENTICATION) {
+            Credentials credentials = extractCredentials(data);
+            onAuthentication(credentials);
+        } else if (event == LockEvent.CANCELED) {
+            onError(new AuthenticationException("a0.authentication_canceled", "The user pressed back"));
```
lbalmaceda, line 60: If the event is "canceled", then raise an exception with the specific code that represents this. This is only for `AuthenticationException`. `LockCallback` would still receive this event and is up to the dev to handle (this advanced scenario remains available)

lib/src/main/java/com/auth0/android/lock/Lock.java
```diff
@@ -168,7 +167,7 @@ private void processEvent(@NonNull Intent data) {
                 break;
             case Constants.INVALID_CONFIGURATION_ACTION:
```
lbalmaceda, line 168: We don't have a "public" event type for this. So we raise an exception instead.

lib/src/main/java/com/auth0/android/lock/LockCallback.java
```diff
@@ -68,5 +69,5 @@
      *
      * @param error describing what happened.
      */
-    void onError(@NonNull LockException error);
+    void onError(@NonNull AuthenticationException error);
```
lbalmaceda, line 72: Changes the type of exception. Helps to deliver the actual exception and not wrapping everything again.

MIGRATION_GUIDE.md
```diff
@@ -38,11 +38,54 @@ As part of removing legacy APIs or authentication flows no longer recommended fo
...
+```
+
 ### Removed classes
 - `VoidCallback` is no longer available. Please, use `Callback<Void, AuthenticationException>` instead.
+- `LockException` is no longer available. This impacts the `LockCallback` and `AuthenticationCallback` classes. Please, use `AuthenticationException` instead.
 
 ### Removed methods
 
+#### From class `AuthenticationCallback`
+- Removed `public void onCanceled()`. Instead, an exception will be raised with this code. Check the property `AuthenticationException#isCanceled()` to handle it.
```
Widcket, line 87: Says "with this code" but doesn't include the code.

Widcket (dismissed):
LGTM, just have a question regarding the Readme.

MIGRATION_GUIDE.md
lbalmaceda: Good catch. Rephrased this and also noted that this is a method and not a property. Ready for review.

Widcket (approved):

PR #622, opened 2021-04-30
Enable AppLinks in the declared intent filters

lbalmaceda:
### Changes

This aligns with what the SDK does. By setting the `autoVerify` attribute to true, and only when the scheme used is "https", upon install time, the Android OS will attempt to pull the _assetlinks_ hosted file from the Auth0 server in order to verify that the domain given in the intent-filters is owned by the developer of the android app. When this succeeds, the disambiguation dialog will not be shown, and the application will be opened directly. This only works for Android 23 and above. 

Having this attribute enabled for when the scheme is not "https" or when the android version is lower than 23, has no negative effect.

Sample logging in with Passwordless Link
2021-04-30 at 11 32 35 - Salmon Pony

Tested running the following command in the terminal
```
adb shell am start -a android.intent.action.VIEW -d "demo://lbalmaceda.auth0.com/android/com.auth0.android.lock.app/email\?code\=145604"
```

See https://github.com/auth0/Lock.Android/pull/620

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,14 +55,6 @@ If you haven't done yet, go to [Auth0](https://auth0.com) and create an Account,
 https://{YOUR_AUTH0_DOMAIN}/android/{YOUR_APP_PACKAGE_NAME}/callback
 ```
 
-If you plan to use Passwordless authentication with SMS or Email connections, the Callback URL you will register looks like this:
-
-```
-https://{YOUR_AUTH0_DOMAIN}/android/{YOUR_APP_PACKAGE_NAME}/sms
-// or
-https://{YOUR_AUTH0_DOMAIN}/android/{YOUR_APP_PACKAGE_NAME}/email
-``` 
-
 The *package name* value required in the Callback URL can be found in your app's `build.gradle` file in the `applicationId` property. Both the *domain* and *client id* values can be found at the top of your Auth0 Application's settings. You're going to use them to setup the SDK. It's good practice to add them to the `strings.xml` file as string resources that you can reference later from the code. This guide will follow that practice.
 
 ```xml
--- a/lib/src/main/AndroidManifest.xml
+++ b/lib/src/main/AndroidManifest.xml
@@ -9,8 +9,8 @@
         <activity
             android:name="com.auth0.android.provider.RedirectActivity"
             tools:node="replace">
-            <intent-filter>
-                <action android:name="android.intent.action.VIEW" />
+            <intent-filter android:autoVerify="true">
+            <action android:name="android.intent.action.VIEW" />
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -55,14 +55,6 @@ If you haven't done yet, go to [Auth0](https://auth0.com) and create an Account,
 https://{YOUR_AUTH0_DOMAIN}/android/{YOUR_APP_PACKAGE_NAME}/callback
 ```
 
-If you plan to use Passwordless authentication with SMS or Email connections, the Callback URL you will register looks like this:
-
-```
-https://{YOUR_AUTH0_DOMAIN}/android/{YOUR_APP_PACKAGE_NAME}/sms
```
lbalmaceda, line 61: when you start a passwordless flow and ask for "android_link" this is the format of callback URL that the server will use. This is not changeable by the developer. 

The server takes the domain from the Auth0 application and the Android package name from this setting:
image

lib/src/main/AndroidManifest.xml
```diff
@@ -9,8 +9,8 @@
         <activity
             android:name="com.auth0.android.provider.RedirectActivity"
             tools:node="replace">
-            <intent-filter>
-                <action android:name="android.intent.action.VIEW" />
+            <intent-filter android:autoVerify="true">
```
lbalmaceda, line 12: Even if inherited from the SDK declaration of this same activity, I rather am sure I'm setting it to true.

lib/src/main/AndroidManifest.xml
```diff
@@ -37,21 +37,21 @@
             android:launchMode="singleTask"
             android:screenOrientation="portrait"
             android:theme="@style/Lock.Theme">
-            <intent-filter>
-                <action android:name="android.intent.action.VIEW" />
+            <intent-filter android:autoVerify="true">
```
lbalmaceda, line 40: Same for passwordless activities, this affects LINK mode

lib/src/main/AndroidManifest.xml
```diff
@@ -37,21 +37,21 @@
...
+            <action android:name="android.intent.action.VIEW" />
 
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
 
                 <data
                     android:host="${auth0Domain}"
                     android:pathPrefix="/android/${applicationId}/email"
-                    android:scheme="${auth0Scheme}" />
+                    android:scheme="https" />
```
lbalmaceda, line 49: the scheme is not customizable anywhere (API nor dashboard), so there's no reason to dynamically set one.

evansims (approved):

PR #623, opened 2021-05-04
Fix filtering by Country name [SDK-2546]

lbalmaceda:
### Changes
The filter was broken leading to an empty list of countries. This PR fixes the filter and makes the list display all the countries at first and then update as the search box changes. 

#### Before:
Green Chicken

#### After:
Blue Sailfish

### References
See SDK-2546

### Testing

Manually tested, as seen on the gif.

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -133,7 +133,7 @@ In case you are not currently calling it, make sure to update your code adding t
 ```kotlin
 class MyActivity : AppCompatActivity() {
 
-  private var lock: Lock? = null
+  private lateinit var lock: Lock
 
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
@@ -148,14 +148,14 @@ class MyActivity : AppCompatActivity() {
 
   private fun launchLock() {
     // Invoke as many times as needed
-    val intent = lock!!.newIntent(this)
+    val intent = lock.newIntent(this)
     startActivity(intent)
   }
  
   override fun onDestroy() {
       super.onDestroy()
       // Release Lock resources
-      lock?.onDestroy(this)
+      lock.onDestroy(this)
   }
 }
 ```
--- a/lib/src/main/java/com/auth0/android/lock/CountryCodeActivity.java
+++ b/lib/src/main/java/com/auth0/android/lock/CountryCodeActivity.java
@@ -49,25 +49,23 @@ package com.auth0.android.lock;
 
 
 import android.content.Intent;
-import android.os.AsyncTask;
 import android.os.Bundle;
-import androidx.annotation.Nullable;
-import androidx.appcompat.app.ActionBar;
-import androidx.appcompat.app.AppCompatActivity;
 import android.text.Editable;
 import android.text.TextWatcher;
 import android.util.Log;
 import android.widget.EditText;
 import android.widget.ListView;
 
+import androidx.annotation.Nullable;
+import androidx.appcompat.app.ActionBar;
+import androidx.appcompat.app.AppCompatActivity;
+
 import com.auth0.android.lock.adapters.Country;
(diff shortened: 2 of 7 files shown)


lib/src/main/java/com/auth0/android/lock/utils/LoadCountriesTask.java
```diff
@@ -26,53 +26,48 @@
...
+import java.util.List;
 import java.util.Map;
 
-public abstract class LoadCountriesTask extends AsyncTask<String, Void, Map<String, String>> {
+public abstract class LoadCountriesTask extends AsyncTask<Context, Void, List<Country>> {
 
     private static final String TAG = LoadCountriesTask.class.getName();
-    public static final String COUNTRIES_JSON_FILE = "com_auth0_lock_passwordless_countries.json";
-
-    private final ThreadLocal<Context> context = new ThreadLocal<>();
```
lbalmaceda, line 48: this reference changed at some point, making the task not to execute at all. I'm passing this as a "task param" rather than via the constructor

lib/src/main/java/com/auth0/android/lock/utils/LoadCountriesTask.java
```diff
@@ -26,53 +26,48 @@
...
         try {
-            final Reader reader = new InputStreamReader(ctx.getAssets().open(params[0]));
+            final Reader reader = new InputStreamReader(params[0].getAssets().open(COUNTRIES_JSON_FILE));
             codes = new Gson().fromJson(reader, mapType);
             Log.d(TAG, String.format("Loaded %d countries", codes.size()));
         } catch (IOException e) {
             Log.e(TAG, "Failed to load the countries list from the JSON file", e);
         }
-        return codes;
+        final ArrayList<String> names = new ArrayList<>(codes.keySet());
```
lbalmaceda, line 64: including the sorting logic in the background task, returning a list of countries instead of a map of strings..

lib/src/main/java/com/auth0/android/lock/adapters/CountryAdapter.java
```diff
@@ -114,12 +112,8 @@ protected FilterResults performFiltering(CharSequence constraint) {
         @Override
         @SuppressWarnings("unchecked")
         protected void publishResults(CharSequence constraint, FilterResults results) {
-            data = (List<Country>) results.values;
-            if (results.count > 0) {
-                notifyDataSetChanged();
-            } else {
-                notifyDataSetInvalidated();
```
lbalmaceda, line 121: docs about this method say that once called, the adapter is no longer valid. https://developer.android.com/reference/android/widget/BaseAdapter#notifyDataSetInvalidated()

lib/src/main/java/com/auth0/android/lock/CountryCodeActivity.java
```diff
@@ -94,48 +101,31 @@ public void beforeTextChanged(CharSequence s, int start, int count, int after) {
...
-
-        task = new LoadCountriesTask(this) {
-            @Override
-            protected void onPostExecute(Map<String, String> result) {
-                task = null;
-                if (result == null) {
-                    return;
-                }
-
-                final ArrayList<String> names = new ArrayList<>(result.keySet());
```
lbalmaceda, line 120: moved into the task's responsibilities

MIGRATION_GUIDE.md
```diff
@@ -133,7 +133,7 @@ In case you are not currently calling it, make sure to update your code adding t
 ```kotlin
 class MyActivity : AppCompatActivity() {
 
-  private var lock: Lock? = null
+  private lateinit var lock: Lock
```
lbalmaceda, line 136: A simpler syntax for Android apps. "lateinit" means that the variable will be initialized later (e.g. in the on create method). Prevents us from asserting the null scenarios.

jimmyjames (approved):

PR #645, opened 2022-06-22
Fixing bug where custom audience was not being set in

Blake-Carrier:
passwordless login

### Changes

Please describe both what is changing and why this is important. Include:

- Added a check for custom audience in the PasswordlessLockActivity
- This fixes a bug where the custom audience was not being passed to Auth0 in a passwordless lock builder scenario

### Testing

- [ x] This change has been tested on the latest version of the platform/language


poovamraj (approved):
👍 LGTM! Thanks for the contribution.

